PR #6, opened 2021-03-28
Limiting the Search Space in Optimal Quantum Circuit Mapping

IsFairy:
This PR adds the functionality to limit the search space during optimal circuit mapping introduced in:

L. Burgholzer, S. Schneider, and R. Wille. *Limiting the Search Space in Optimal Quantum Circuit Mapping*. In *Asia and South Pacific Design Automation Conference (ASP-DAC)*, 2022.

Furthermore, it brings the following additions and changes:
 - [x] ⚡ actually enables -march=native -mtune=native for source builds via pip
 - [x] ⚡ setup.py now runs on as many threads as available
 - [x] 🏗️ switch to PEP 517 `pyproject.toml`
 - [x] 💚 switch cibuildwheel configuration to `pyproject.toml`
 - [x] 🚀 added Python 3.10 wheels
 - [x] ♻️ refactor the Python bindings to make them easier to use
 - [x] ⬆️ updated all external submodules to their most recent version
 - [x] 📝 updated documentation


src/encodings/Encodings.cpp
```diff
@@ -0,0 +1,389 @@
...
+	if (connections.at(node).empty()){
+		visited[node]=false;
+		return;
+	}
+
+	for (auto child: connections.at(node)){
+		findLongestPath(child, curSum+1);
+	}
+
+	visited[node]=false;
```
burgholzer, line 388: This should not be necessary. The Architecture class already instantiates a `distanceTable` at construction time which represents a Dijkstra tableau of the distances between all the nodes. It can be accessed via `architecture.getDistanceTable()` and the longest path is then given by the maximum over the 2D array.
Consequently the `Architecture` class should have a `getLongestPathRoutine` which does exactly this.

src/exact/ExactMapper.cpp
```diff
@@ -59,6 +59,8 @@ void ExactMapper::map(const MappingSettings& settings) {
 	} while (next_combination(qubits.begin(), qubits.begin()+qc.getNqubits(), qubits.end()));
 
 	// 3) determine exact mapping for this qubit choice
+	if (this->settings.bddLimits>0)	
+		this->settings.bddLimits = findLongestPath(architecture.getCouplingMap(), architecture.getNqubits());
```
burgholzer, line 63: there should be a flag that enables/disables this feature and a separate variable indicating the limit.

src/exact/ExactMapper.cpp
```diff
@@ -286,26 +293,64 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
...
+				}
+				if (grouping > 0) {
+					opt.add(exactlyOneCMDR(varIDs, groupVars(varIDs, grouping), (auxvars.size() - 1), auxvars, c));
+				}
+				else if (grouping == -1) {
+					opt.add(exactlyOneCMDR(varIDs, groupVars(varIDs, log(varIDs.size())), (auxvars.size() - 1), auxvars, c));
+				}				
+                else if (grouping == -2) {
+					opt.add(exactlyOneCMDR(varIDs, groupVars(varIDs, varIDs.size() / 2), (auxvars.size() - 1), auxvars, c));
+				}
```
burgholzer, line 348: enums for grouping and encoding

src/exact/ExactMapper.cpp
```diff
@@ -359,21 +404,43 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
...
+			}
+			else if (grouping == -1) {
+				opt.add(exactlyOneCMDR(varIDs, groupVars(varIDs, log(varIDs.size())), -1, auxvars, c));
+			}			
+            else if (grouping == -2) {
+				opt.add(exactlyOneCMDR(varIDs, groupVars(varIDs, varIDs.size() / 2), -1, auxvars, c));
+			}
 
+		}
+	}
```
burgholzer, line 437: enums for grouping and encoding

test/test_exact.cpp
```diff
@@ -87,3 +87,67 @@ TEST_P(ExactTest, QubitTriangle) {
...
+}
+
+TEST_P(ExactTest, BDDLimits) {
+	MappingSettings settings{};
+	settings.bddLimits = 1;
+	IBMQ_Yorktown_mapper.map(settings);
+	IBMQ_Yorktown_mapper.dumpResult(GetParam() + "_exact_yorktown_commander.qasm");
+	IBMQ_Yorktown_mapper.printResult(std::cout, true);
+	SUCCEED() << "Mapping successful";
+}
```
burgholzer, line 153: These tests do not cover all of the changes in this PR, e.g., they only test on a single bidirectional architecture.
Consider writing more tests. 
(CI should pass again now)

burgholzer (changes requested):
This needs some work in order to be merged into master. See the comments for details.

include/encodings/Encodings.hpp
```diff
@@ -0,0 +1,86 @@
...
+inline
+bool operator==(const WeightedVar& rhs, const WeightedVar& lhs) {
+	return rhs.weight == lhs.weight && rhs.varID == lhs.varID;
+}
+
+struct SavedLit
+{
+	SavedLit() :type(-1), varID(0) {}
+	SavedLit(int type, unsigned long varID) : type(type), varID(varID) {}
+	int type = -1;
```
burgholzer, line 48: Wouldn't it be more explanatory if `type` was not merely an `int` but an inner enum class `Type` with its possible values? Something like:
```suggestion
        enum class Type { neg, pos } // whatever the possible types are
	Type type = Type::pos; // or whatever the default shall be
```

burgholzer (changes requested):
One minor comment, see above.

There are lots of unused functions in the `Encodings.cpp`. Is this intentional? If not, consider removing them.

Is there a reason for the `Encoding.hpp/.cpp` files being in a new directory instead of the top-level `include`/`src`? I wouldn't expect such a directory to grow with additional files in the future.

What's still left to do is to add the new commandline options to the Python Bindings in `jkq/qmap/bindings.cpp` and update the README.

burgholzer (commented):
The settings used for mapping should be somehow reflected in the `MappingResults` struct as with the other statistics (e.g., `layeringStrategy`).
I'd suppose `toString` methods are needed as for the other `enum class` settings to automatically translate an enum value to its corresponding string in the resulting JSON object.
I am open for suggestions to handle this in a more efficient manner.

apps/exact_app.cpp
```diff
@@ -139,7 +139,7 @@ int main(int argc, char** argv) {
 				ms.bddStrategy = BDDStrategy::Custom;
 				if (vm.count("bdd_limits")) {
 					const std::string bdd_limits = vm["bdd_limits"].as<std::string>();
-					ms.bddLimits = std::stoi(bdd_limits.c_str());
+					ms.bddLimit = std::stoi(bdd_limits.c_str());
```
burgholzer: Once instance of `bdd_limits` with an `s` left here.

apps/exact_app.cpp
```diff
@@ -87,6 +92,66 @@ int main(int argc, char** argv) {
...
+	if (vm.count("bdd"))
+	{
+		ms.enableBDDLimits = true;
+		if (vm.count("bddStrategy")) {
+			const std::string bddStrat = vm["bddStrategy"].as<std::string>();
+			if (bddStrat == "custom") {
+				ms.bddStrategy = BDDStrategy::Custom;
+				if (vm.count("bdd_limits")) {
+					const std::string bdd_limits = vm["bdd_limits"].as<std::string>();
+					ms.bddLimit = std::stoi(bdd_limits.c_str());
```
burgholzer, line 142:
```suggestion
				if (vm.count("bdd_limit")) {
					const std::string bdd_limit = vm["bdd_limit"].as<std::string>();
					ms.bddLimit = std::stoi(bdd_limit.c_str());
```

README.md
```diff
@@ -92,6 +92,11 @@ Params:
     use_teleportation - Use teleportation in addition to swaps
     teleportation_fake - Assign qubits as ancillary for teleportation but don't actually use them
     teleportationSeed - Fix a seed for the initial ancilla placement (0 means no fixed seed)
+    encoding - Choose encoding for AMO and exactly one (none | commander | bimander)
+	grouping - Choose method of grouping (fixed2 | fixed3 | logarithm | halves)
+	bdd - Enable bdd for limiting swaps per layer
+	bddStrategy - Choose method of applying bdd limits (none | custom | architectureswaps | subsetswaps)
+    bdd_limit - Set a custom limit for max swaps per layer
```
burgholzer, line 99: these are not in the function signature below.

src/exact/ExactMapper.cpp
```diff
@@ -359,21 +433,42 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
...
+		for (unsigned long k = 1; k < reducedLayerIndices.size(); ++k) {
+			expr onlyOne = c.int_val(0);
+			piCount = 0;
+			do {
+				onlyOne = onlyOne + ite(y[k - 1][piCount], c.int_val(1), c.int_val(0));
+				++piCount;
+			} while (std::next_permutation(pi.begin(), pi.end()));
+			opt.add(onlyOne.simplify() == 1);
+		}
+	} else if (this->settings.encoding != Encodings::None) {
```
burgholzer, line 446:
```suggestion
	} else {
```

README.md
burgholzer: also, please indicate the intended default values by surrounding them with asterisks.

burgholzer (changes requested):
A few minor things before this can be merged 👍🏼 

The `Command-line Executable` section in the README also needs to be updated to describe the new options for the exact mapper.

`jkq/qcec/compile.py` needs to be updated to accommodate the new options or the Python bindings won't work.

src/Encodings.cpp
```diff
@@ -0,0 +1,322 @@
+
+#include "Encodings.hpp"
```
burgholzer, line 2: Lots of lines not covered by tests. Consider adding some tests for these cases.

burgholzer (changes requested):
Overall this looks good. I have some comments and questions, mostly related to sensible defaults for all the new options. From a user perspective I would imagine that I can pick between the following high-level options:
 - useSubsets (default *true*)
 - limitSwaps (default "true")

More of a general question: Does it make sense to include the BDD method? Is there a true benefit of using it?

Some of the lines not covered by tests should be easy to cover.

README.md
```diff
@@ -163,6 +175,25 @@ Both, the exact and the heuristic mapping tool also offer the `--layering` optio
...
+As commander encoding can use different strategies to group the variables, there are different `--grouping` options:
+- `halves` (*default*): each group contains half of the total variables
+- `logarithm`: each group contains at most log2 of the total variables
+- `fixed2`: each group contains exactly two variables
+- `fixed3`: each group contains exactly three variables
+
+The exact mapping tool also offers the `--strategy` option to enable limiting the number of swaps done per layer, which offers the following options:
+- `none` (*default*): disable bdd limiting
+- `architectureswaps`: calculate the max swaps per layer based on the longest path through the whole coupling map
+- `subsetswaps`: calculate the max swaps per layer based on the longest path of current choice of qubits
```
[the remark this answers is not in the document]
IsFairy: true, I'm also renaming this to something else, longestpath maybe

include/MappingSettings.hpp
```diff
@@ -1,95 +1,182 @@
...
+    bool verbose             = false;
+
+    bool               lookahead            = true;
+    int                nrLookaheads         = 15;
+    int                teleportationQubits  = 0;
+    unsigned long long teleportationSeed    = 0;
+    bool               teleportationFake    = false;
+    double             firstLookaheadFactor = 0.75;
+    double             lookaheadFactor      = 0.5;
+    Encodings          encoding             = Encodings::None;
```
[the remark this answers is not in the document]
IsFairy: not in general, while the encodings are faster, esp. for larger circuits, smaller circuits actually are faster without the overhead

README.md
```diff
@@ -163,6 +175,25 @@ Both, the exact and the heuristic mapping tool also offer the `--layering` optio
...
+- `increasing`: start with 0 swaps and geometrically increase the number of swaps per layer
+- `custom`: set a custom limit, needs the `--limit` option followed by a number to set the limit
+Using the `--useBDD` option, the mapping utilizes bdds instead of simply removing the permutations from the core routine.
```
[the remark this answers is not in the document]
burgholzer: Yap. Sounds great. Just from a user perspective it should be clear that this is an option that could be worth trying out but is not guaranteed to perform any better.

include/MappingSettings.hpp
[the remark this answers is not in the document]
burgholzer: Is there some sort of "threshold" for when it makes sense to switch to the encodings?
Is it small circuits with respect to the number of qubits or rather small circuits regarding the gate count?
It would be great to have some sensible defaults that are chosen automatically (maybe even depending on the benchmark)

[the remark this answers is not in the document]
IsFairy: both actually, but the qubit count is the more important.
I don't really like when programs choose options for me automatically, other than a given default that works in the majority of cases, so that is something I always try to avoid (because when I'm the user, I then often wonder why suddenly the config changes and have to dig through the documentation anyway)

src/exact/ExactMapper.cpp
```diff
@@ -1,486 +1,658 @@
...
+    int runs = 1;
+    for (auto& choice: allPossibleQubitChoices) {
+        std::size_t limit      = 0;
+        std::size_t upperLimit = this->settings.limit;
+        if (this->settings.strategy == Strategy::ArchitectureSwaps) {
+            limit = this->architecture.getLongestPath() - 1;
+        } else if (this->settings.strategy == Strategy::SubsetSwaps) {
+            limit = this->architecture.getLongestPath(choice) - 1;
+        } else if (this->settings.strategy == Strategy::Increasing) {
+            ///  TODO: is there anything supposed to happen here?
```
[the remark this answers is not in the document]
IsFairy: this actually can be removed, the code that does the work for increasing strategy is at the end of the do while loop, increasing the limit and stopping when an upper limit is reached..

PR #8, opened 2021-04-06
Fix for teleportation

### LINKED ISSUE #7: Teleportation sometimes fails
The teleportation functionality is flaky and fails with some initial placements.

hillmich:
The previous code was not ported correctly. Fixes #7

The ancilla selection was wrong and so was the case where not suitable edge was found in `mapUnmappedGates`.

### DIFF
--- a/include/Architecture.hpp
+++ b/include/Architecture.hpp
@@ -53,6 +53,7 @@ class Architecture {
 	static constexpr bool VERBOSE = false;
 
 public:
+    std::vector<std::pair<short, short>> teleportation_qubits{};
 
 	struct CalibrationData {
 		unsigned short qubit = 0;
--- a/include/Architecture.hpp
+++ b/include/Architecture.hpp
@@ -53,9 +53,8 @@ class Architecture {
 	static constexpr bool VERBOSE = false;
 
 public:
-    std::vector<std::pair<short, short>> teleportation_qubits{};
 
-	struct CalibrationData {
+    struct CalibrationData {
 		unsigned short qubit = 0;
 		double t1 = 0.0; // [ms]
 		double t2 = 0.0; // [ms]
@@ -106,6 +105,9 @@ public:
 	CouplingMap& getCurrentTeleportations() {
 		return current_teleportations;
 	}
+    std::vector<std::pair<short, short>>& getTeleportationQubits() {
+        return teleportationQubits;
+    }
 
 	const Matrix& getDistanceTable() const {
 		return distanceTable;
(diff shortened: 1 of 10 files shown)


include/Architecture.hpp
```diff
@@ -53,6 +53,7 @@ class Architecture {
 	static constexpr bool VERBOSE = false;
 
 public:
+    std::vector<std::pair<short, short>> teleportation_qubits{};
```
burgholzer, line 56: Very minor: shouldn't this be named `teleportationQubits` in order to conform with (most of) the other names?
This also holds for the identically named member in the `MappingSettings` class.

src/heuristic/HeuristicMapper.cpp
```diff
@@ -302,19 +301,40 @@ void HeuristicMapper::mapUnmappedGates(long layer, HeuristicMapper::Node& node,
...
+                // TODO: Consider fidelity here if available. The best available edge should be chosen
+                auto& chosenEdge = *possibleEdges.begin();
+                locations.at(gate.control) = chosenEdge.first;
+                locations.at(gate.target) = chosenEdge.second;
+                qubits.at(chosenEdge.first) = gate.control;
+                qubits.at(chosenEdge.second) = gate.target;
+                qc::QuantumComputation::findAndSWAP(gate.control, chosenEdge.first, qcMapped.initialLayout);
+                qc::QuantumComputation::findAndSWAP(gate.target, chosenEdge.second, qcMapped.initialLayout);
+                qc::QuantumComputation::findAndSWAP(gate.control, chosenEdge.first, qcMapped.outputPermutation);
+                qc::QuantumComputation::findAndSWAP(gate.target, chosenEdge.second, qcMapped.outputPermutation);
```
burgholzer, line 336: duplicate code (from other branch of the if statement).
could be removed.

src/heuristic/HeuristicMapper.cpp
```diff
@@ -302,19 +301,40 @@ void HeuristicMapper::mapUnmappedGates(long layer, HeuristicMapper::Node& node,
...
+                    for (int j = i + 1; j < architecture.getNqubits(); j++) {
+                        if (qubits.at(i) == DEFAULT_POSITION && qubits.at(j) == DEFAULT_POSITION) {
+                            double dist = architecture.distance(i, j);
+                            if (dist < bestScore) {
+                                bestScore = dist;
+                                bestMapping = std::make_pair(i, j);
+                            }
+                        }
+                    }
+                }
```
burgholzer, line 316: Could this possibly be a member function of the `Architecture` class?
More generally: would it make sense to add functions to the architecture class that return the shortest/longest path?
Of course this is out of scope for this PR, so feel free to ignore it for now.

burgholzer (commented):
Some very minor remarks. LGTM otherwise 👍🏼

src/heuristic/HeuristicMapper.cpp
hillmich: Yeah, this could be moved to the Architecture class. However, as we currently only use this at this place, I'd do this as refactor later as this is too specific for general usage at the moment.

burgholzer (approved):
LGTM 👍🏼

PR #14, opened 2021-11-29
Portable wheels for Apple Silicon

burgholzer:
Up until now, building portable wheels for Apple Silicon did not work properly. This is due to our setup not producing binaries for `arm64`. This PR changes that by allowing to infer the architecture from the environment in `setup.py`.
The external z3 dependency is resolved by downloading the corresponding Apple Silicon bottle from `brew` and installing it manually.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ defaults:
 env:
   BUILD_TYPE: Release
   CMAKE_BUILD_PARALLEL_LEVEL: 2
-  Z3_GIT_TAG: z3-4.8.10
+  Z3_GIT_TAG: z3-4.8.14
 
 jobs:
   codestyle:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,7 +10,9 @@ on:
   workflow_dispatch:
 
 env:
-  Z3_GIT_TAG: z3-4.8.13
+  Z3_GIT_TAG:   z3-4.8.14
+  Z3_HASH:      884f1128a33ea7dea3ad375e50ef5ea36b26623e2999119cf31252dd4650692f
+  CIBW_VERSION: v2.3.1
 
 jobs:
   build_manylinux_wheels:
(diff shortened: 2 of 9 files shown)


katringoogoo:
hm a different option for this would be to build x86_64 and arm64 separately and don't use the universal2 for now. it seems like homebrew won't adopt universal binaries for the foreseeable future but it should be possible to just install the right z3 package via homebrew even when cross-compiling (in this case install the m1 packages in the x86 macos runner): https://stackoverflow.com/questions/64882584/how-to-run-the-homebrew-installer-under-rosetta-2-on-m1-macbook

the only other option i currently see would be to build a universal z3 lib yourself (like you already do on windows)

katringoogoo:
just to not loose this snippet - one could e.g. parse the archflags and inject the osx architecture this way. in qfr an additional mapping from 'universal2' to 'x86_64;arm64' is needed

setup.py
```python
        # cross-compile support for macOS - respect ARCHFLAGS if set
        if sys.platform.startswith("darwin"):            
            archs = re.findall(r"-arch (\S+)", os.environ.get("ARCHFLAGS", ""))
            if archs:
                arch_argument = "-DCMAKE_OSX_ARCHITECTURES={}".format(";".join(archs))
                print('macOS building with: ', arch_argument, flush=True)
                cmake_args += [arch_argument]
```

burgholzer:
I am all for dropping the `universal2` wheels in favour of building the architectures separately. In the near future there will be support for Apple Silicon GitHub Actions runners, so cross compilation shouldn't be an issue in the long run.
The snippet you posted seems like a great idea. Could/should probably be adopted in the QFR repository as well.

I guess building a universal Z3 binary is quite a hassle and I am not sure whether this is the most efficient way to proceed. (we only do that under Windows because I didn't bother to find out where pip installs all the packages under Windows. Otherwise, windows could be handled similarly to linux)

If there is an efficient way to obtain the M1 Z3 binary from brew for the Apple Silicon wheels during the CI runs, i guess this should be the solution to go for.

hillmich (approved):
LGTM

PR #16, opened 2021-12-29
Compile flag fixes for latest clang and Z3 dependency reduction

katringoogoo:
- Removed Z3 dependencies from the heuristic lib
- Added the ability to build python bindings without Z3
- Fixed compile flags for clang on apple silicon

also see https://github.com/iic-jku/qmap/pull/15 for the original PR


burgholzer:
The reason why macOS bindings fail to build in the CI is most certainly because of: https://github.com/iic-jku/qfr/blob/fbd4310d7e7c8b88b314376283dcb15b61e5f3ad/CMakeLists.txt#L24-L30

If not set explicitly in the top level project, CMake will try to build fat binaries. Unfortunately, Z3 offers no universal wheels for download (which is why #14 is on hold until either Z3 publishes universal wheels or GitHub Actions provides M1 runners). Something like
```cmake
if (DEPLOY)
	set(CMAKE_OSX_ARCHITECTURES "x86_64" CACHE STRING "" FORCE)
endif ()
```
should resolve the error in the meantime.

katringoogoo:
I added the option but i'm not sure this will solve the problem. As far as i can see the qfr project already sets the CMAKE_OSX_ARCHITECTURES and forces it back to 'x86_64;arm64' which in turn breaks delocate-listdeps (it does not expect universal binaries) and i'm not sure about its status either :D https://github.com/matthew-brett/delocate/issues/78

burgholzer:
Hm. I would have guessed that setting this before including the submodule overrides anything done in the submodule. Does not seem like it 🙃 
Let's continue the discussion in #14 and see whether we can find a solution there that also solves the problem here.

burgholzer:
Now that #14 is merged, this PR should be good to go after a rebase 👍🏼

PR #15, opened 2021-12-28
Compile flag fixes for latest clang and Z3 dependency reduction

katringoogoo:
- Removed Z3 dependencies from the heuristic lib
- Added the ability to build python bindings without Z3
- Fixed compile flags for clang on apple silicon

PR #21, opened 2022-01-07
Compile flag and z3 dependency fixes

katringoogoo:
- Removed Z3 dependencies from the heuristic lib
- Added the ability to build python bindings without Z3
- Fixed compile flags for clang on apple silicon

Recreated PR on new master after fixes (also see #15 and #16 for reference)


burgholzer:
Thanks again for the PR. Could you merge the PR with the current master one last time?
I just found out that a stray `.codecov.yaml` somehow remained in the root directory of the project which contained outdated settings for code coverage.
Merging the change should also fix the failing test. 🙃

katringoogoo:
sure :)

burgholzer (approved):

PR #78, opened 2022-06-27
Clifford synthesis

IsFairy:
Adding the functionality to synthesize clifford circuits from tableaus and unoptimized input circuits, with some structural changes.

First, Adding LogicBlocks as a dependency and the logging functionality from there.
Introduction of Cliffordresults class to contain all Tableaus and Resulting Circuits.
Introduction of Tableaus as data structure to represent Clifford circuits.

ToDos:

- [x] Check integration of Logicblocks for sneaky errors 
- [x] Check functionality of Cliffordsynthesiszer
- [x] Write Tests for new Classes
- [x] Add ability to load a tableau from a tableau string as in qiskit (https://qiskit.org/documentation/stubs/qiskit.quantum_info.Clifford.html)
- [x] Add Python bindings


src/Tableau.cpp
```diff
@@ -0,0 +1,395 @@
...
+        {
+            if (gate->isControlled()) {
+                util::fatal("Expected single-qubit gate");
+            }
+            const auto a = gate->getTargets().at(0U);
+            for (auto i = 0U; i < nqubits; i++) {
+                tableau[i][2U * nqubits] ^= (tableau[i][a] & tableau[i][a + nqubits]);
+                std::swap(tableau[i][a], tableau[i][a + nqubits]);
+            }
+            return 1U;
```
burgholzer, line 113: please extract these updated into separate functions. there is so much code duplication in this whole function.

src/Tableau.cpp
```diff
@@ -0,0 +1,395 @@
...
+                tableau[i][a + nqubits] ^= tableau[i][a];
+            }
+            for (auto i = 0U; i < nqubits; i++) {
+                tableau[i][2U * nqubits] ^= tableau[i][a] & tableau[i][a + nqubits];
+                tableau[i][a + nqubits] ^= tableau[i][a];
+            }
+            return 3U;
+        }
+        case qc::OpType::Z: { // Z = S x S
+            if (gate->isControlled()) {
```
burgholzer, line 187: it should be pretty easy to support controlled-Z gates since it's just a CNOT and two Hadamards.

src/Tableau.cpp
```diff
@@ -0,0 +1,395 @@
...
+            for (auto i = 0U; i < nqubits; i++) {
+                tableau[i][2U * nqubits] ^= tableau[i][a] & tableau[i][a + nqubits];
+                tableau[i][a + nqubits] ^= tableau[i][a];
+            }
+            return 2U;
+        }
+        case qc::OpType::Y: { // Y = H x S x S x H x S x S
+            if (gate->isControlled()) {
+                util::fatal("Expected single-qubit gate");
+            }
```
burgholzer, line 204: as above, supporting controlled-Y is easy since it's a CNOT sandwiched by an Sdag and a S on the target.

src/Tableau.cpp
```diff
@@ -0,0 +1,395 @@
...
+    }
+}
+
+void Tableau::initTableau(Tableau& tableau, std::size_t nqubits) {
+    tableau.init(nqubits);
+}
+
+int Tableau::applyGate(const std::unique_ptr<qc::Operation>& gate) {
+    auto nqubits = getQubitCount();
+    switch (gate->getType()) {
```
burgholzer, line 102: this whole list is missing the SWAP gate.

src/Tableau.cpp
```diff
@@ -0,0 +1,395 @@
...
+    TableauType result{};
+    result.resize(nQubits);
+    for (auto i = 0; i < nQubits; i++) {
+        result[i].resize(2U * nQubits + 1U);
+        for (auto j = 0; j < 2 * nQubits; j++) {
+            if (i == j - nQubits) {
+                result[i][j] = 1;
+            } else {
+                result[i][j] = 0;
+            }
```
burgholzer, line 249: this else branch can be dropped since the default initialisation is `0` anyways.

src/Tableau.cpp
```diff
@@ -0,0 +1,395 @@
...
+    } else {
+        for (int i = 0; i < nQubits; ++i) {
+            auto first  = std::find_if(tableau[i].begin(), tableau[i].end(), [](const auto x) { return x == 1; });
+            auto last   = std::find_if(tableau[i].rbegin(), tableau[i].rend(), [](const auto x) { return x == 1; });
+            auto first2 = std::find_if(other.tableau[i].begin(), other.tableau[i].end(), [](const auto x) { return x == 1; });
+            auto last2  = std::find_if(other.tableau[i].rbegin(), other.tableau[i].rend(), [](const auto x) { return x == 1; });
+            auto d1     = std::distance(tableau[i].begin(), first);
+            auto d2     = std::distance(tableau[i].rbegin(), last);
+            auto d3     = std::distance(other.tableau[i].begin(), first2);
+            auto d4     = std::distance(other.tableau[i].rbegin(), last2);
```
burgholzer, line 270: these should all be const

src/Tableau.cpp
```diff
@@ -0,0 +1,395 @@
...
+                result[i][j] = 0;
+            }
+        }
+        result[i][2U * nQubits] = 0;
+    }
+
+    return Tableau(result);
+}
+
+double Tableau::tableauDistance(Tableau other, int nQubits) {
```
burgholzer, line 257: is this a somehow established distance metric on tableaus?

src/Tableau.cpp
```diff
@@ -0,0 +1,395 @@
...
+            }
+        }
+        if (Tableau::tableauDistance(diagonal.tableau, intermediate_result, nQubits) <
+            Tableau::tableauDistance(diagonal.tableau, result, nQubits)) {
+            result = intermediate_result;
+        }
+    } while (std::next_permutation(indices.begin(), indices.end()));
+    return Tableau(result);
+}
+double Tableau::tableauDistance(const TableauType& tableau1, const TableauType& tableau2, std::size_t nQubits) {
```
burgholzer, line 316: why is this code duplicated here? please unify both functions.

src/Tableau.cpp
```diff
@@ -0,0 +1,395 @@
...
+unsigned long Tableau::getBVFrom(int column) const {
+    unsigned long result = 0UL;
+    for (std::size_t j = 0; j < getQubitCount(); ++j) {
+        if (tableau[j][column] == 1) {
+            result |= (1U << j);
+        }
+    }
+    return result;
+}
+void Tableau::init(std::size_t nQubits) {
```
burgholzer, line 345: again, this is a duplicate fragment of code. try to reuse wherever possible.

burgholzer (changes requested):
This is just a very rough and incomplete review. This PR introduces quite some new functionality, so it might take me a while to get through everything.
In order to not halt the progress, my remarks focus on both ends of the spectrum:
 - the tableau implementation
 - the outside Python interface and functionality

In addition to the inline remarks: we recently introduced some new checks to QMAP that ensure the use of modern Python concepts. It would be great to fix all the stuff reported by pre-commit and/or mypy.
Furthermore, there now is proper documentation hosted at RtD. The Clifford Optimizer should certainly be featured there. Maybe it's worth to wait with this until the public interface is consolidated a bit more. Just a thing to keep in mind.

In case of any question, do not hesitate to reply to the respective comments. 
I'll try to make it through the rest of the PR over the course of the week.

CMakeLists.txt
```diff
@@ -65,6 +65,9 @@ if(NOT Z3_FOUND)
   message(WARNING "Did not find Z3. Exact library and other depending target will not be available")
 endif()
 
+set(THREADS_PREFER_PTHREAD_FLAG ON)
+find_package(Threads REQUIRED)
```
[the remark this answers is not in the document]
IsFairy: The multithreaded variant of CliffordSynthesis needs Threads

src/Tableau.cpp
IsFairy: This is just a metric that worked best for embedding tableaus, in some limited testing with other distance metrics for binary matrices. I'll make a comment noting it

CMakeLists.txt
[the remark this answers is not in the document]
burgholzer: Alright. I am not quite sure whether this is actually required on modern systems anymore (We do use `std::thread` in QCEC and don't link to `Threads::Threads`. Still runs under all tested systems.
Nevertheless, I am probably in favor of keeping this. Feel free to resolve this conversation.

CMakeLists.txt
```diff
@@ -15,6 +15,9 @@ option(COVERAGE "Configure for coverage report generation")
 option(GENERATE_POSITION_INDEPENDENT_CODE "Generate position independent code")
 option(BUILD_QMAP_TESTS "Also build tests for QMAP project")
 
+set(BUILD_QMAP_TESTS ON)
+set(BINDINGS ON)
+
```
burgholzer, line 20: These should be removed before merging.

apps/cliffordsynthesis_app.cpp
```diff
@@ -0,0 +1,269 @@
...
+#include "LogicBlock/LogicBlock.hpp"
+#include "algorithms/RandomCliffordCircuit.hpp"
+#include "cliffordsynthesis/CliffordSynthesizer.hpp"
+#include "utils/logging.hpp"
+
+#include <boost/program_options.hpp>
+#include <iostream>
+#include <locale>
+
+int main(int argc, char** argv) {
```
burgholzer, line 15: This is more of a general question: is it really practical to maintain this CLI in parallel to the Python interface?
At least from my point of view, the future of advertising the usage of this tool (and qmap as a whole) is via the Python interface. It allows far easier connection to things like qiskit and is more flexible. Furthermore, the CLI tools are not available when installing via pip. At least in my opinion, they are a pretty high burden and poorly maintained.
I am in favor of removing this CLI entirely.

examples/cliffordexamples/base-tableau.tabl
```diff
@@ -0,0 +1,3 @@
+2|1|2|3|4|R|
+1|1|1|0|0|0|
+2|0|0|1|1|0|
```
burgholzer, line 3: Is that format in some way established? It's not really self explanatory to me. I would have assumed the most simple format to be a binary table/matrix representing the stabilizer tableau.

examples/cliffordexamples/destabilizer.txt
```diff
@@ -0,0 +1,8 @@
+Destabilizer = ['+IZ', '+XI']
+Destabilizer = ['-XX', '+XI']
+Destabilizer = ['+XX', '+XI']
+Destabilizer = ['+XY', '+XI']
+Destabilizer = ['+XY', '+ZZ']
+Destabilizer = ['+XX', '+ZZ']
+Destabilizer = ['-XX', '+ZZ']
+Destabilizer = ['+XX', '-ZZ']
```
burgholzer, line 8: Is this IBM's format?

examples/cliffordexamples/destabilizer.txt
```diff
@@ -0,0 +1,8 @@
+Destabilizer = ['+IZ', '+XI']
```
burgholzer, line 1: Why "destabilizer" and not stabilizer?

include/Tableau.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+#include <cstdint>
+#include <fstream>
+#include <limits>
+#include <ostream>
+#include <utility>
+#include <vector>
+
+class Tableau {
+    using EntryType   = std::int32_t;
```
burgholzer, line 20: Although this is not the biggest concern right now, but using 32bit for a single entry in a table that is supposed to be binary seems to be very wasteful. Are you actually taking advantage of these 31 bits in any way? If not, then please consider choosing a more suitable type here. Technically, the most performant solution would probably be to use individual bits for every entry. But even `std::uint8_t` already reduces the memory footprint by a factor of 4.

include/cliffordsynthesis/CliffordSynthesizer.hpp
```diff
@@ -0,0 +1,289 @@
...
+        if (optimal_results.resultCircuit.empty()) {
+            std::cerr << "Circuit is empty." << std::endl;
+            return;
+        }
+
+        size_t      dot       = outputFilename.find_last_of('.');
+        std::string extension = outputFilename.substr(dot + 1U);
+        std::transform(extension.begin(), extension.end(), extension.begin(), [](unsigned char c) { return ::tolower(c); });
+        if (extension == "real") {
+            dumpResult(outputFilename, qc::Real);
```
burgholzer, line 81: If I recall correctly, only Qasim export is actually implemented. Accordingly, this can be removed

include/cliffordsynthesis/CliffordSynthesizer.hpp
```diff
@@ -0,0 +1,289 @@
...
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+class CliffordOptimizer {
+public:
+    CliffordOptimizer() = default;
+    void init(bool pchoose_best = false, bool puse_embedding = false, unsigned char pnqubits = 0, unsigned short pinitial_timesteps = 0, SynthesisStrategy strategy = SynthesisStrategy::UseMinimizer, SynthesisTarget ptarget = SynthesisTarget::GATES);
+    void setCircuit(const qc::QuantumComputation& qc);
+    void setTableau(Tableau& targetTabl);
```
burgholzer, line 42:
```suggestion
    void setTargetTableau(Tableau& targetTableau);
```

burgholzer, line 42: I would argue that there should also be a method to set the initial tableau. It might be interesting to also allow different initial states.

include/cliffordsynthesis/CliffordSynthesizer.hpp
```diff
@@ -0,0 +1,289 @@
...
+        H,
+        S,
+        X,
+        Y,
+        Z,
+        Sdag,
+        CX,
+        CY,
+        CZ,
+    };
```
burgholzer, line 173: To me, this class seems rather redundant when looking at the OpType enum. The collections of supported gates could just as well be collections of OpTypes. The `NOP` can just be replaced by the identity gate.

include/cliffordsynthesis/CliffordSynthesizer.hpp
```diff
@@ -0,0 +1,289 @@
...
+#include <thread>
+#include <tuple>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+class CliffordOptimizer {
+public:
+    CliffordOptimizer() = default;
+    void init(bool pchoose_best = false, bool puse_embedding = false, unsigned char pnqubits = 0, unsigned short pinitial_timesteps = 0, SynthesisStrategy strategy = SynthesisStrategy::UseMinimizer, SynthesisTarget ptarget = SynthesisTarget::GATES);
```
burgholzer, line 40: Shouldn't this much rather be part of a constructor?

include/cliffordsynthesis/CliffordSynthesizer.hpp
burgholzer, line 42: In addition to these methods it would actually make sense to provide constructors for the optimizer.
One taking a a circuit + all the configuration options
And one taking a target tableau + all configurations.
Both of these might optionally take an initial tableau as well.
In a similar spirit, they might optionally take an architecture.

include/cliffordsynthesis/CliffordSynthesizer.hpp
```diff
@@ -0,0 +1,289 @@
...
+            const std::set<std::pair<unsigned short, unsigned short>>& reducedCM,
+            const std::vector<unsigned short>& qubitChoice, Tableau& initialTableau,
+            Tableau& targetTableau);
+
+    void make_depth_optimizer(
+            int                                                        timesteps,
+            const std::set<std::pair<unsigned short, unsigned short>>& reducedCM,
+            const std::vector<unsigned short>& qubitChoice, std::unique_ptr<LogicBlock>& lb,
+            const LogicMatrix& x, const LogicMatrix& z, const LogicVector& r,
+            const LogicMatrix3D& g_s, const LogicMatrix3D& g_c) const;
```
burgholzer, line 110: Would it make sense to capture all these logic variables in a `struct`. This would certainly allow for a cleaner interface here.

include/cliffordsynthesis/CliffordSynthesizer.hpp
```diff
@@ -0,0 +1,289 @@
...
+
+protected:
+    Architecture                architecture{};
+    CliffordOptimizationResults main_optimization(
+            int                                                        timesteps,
+            const std::set<std::pair<unsigned short, unsigned short>>& reducedCM,
+            const std::vector<unsigned short>& qubitChoice, Tableau& initialTableau,
+            Tableau& targetTableau);
+
+    void make_depth_optimizer(
```
burgholzer, line 105: Seeing these three functions and the whole code in this file I am wondering. Would it actually make sense to create a base class for the optimizer and then provide three derived classes (in separate files) for the different objective functions. It would certainly clean up the library here and would act as a separation of concerns.

burgholzer (changes requested):
I just had some time to go over more parts of the code. Please find the comments below. Some of these are rather conceptual questions that might need some discussions.

examples/cliffordexamples/destabilizer.txt
IsFairy: Yes, this is IBMs format for stabilizer tables

examples/cliffordexamples/base-tableau.tabl
IsFairy: Not established, but I choose it that way to offer some tradeoff between ease of use in code and readability. Its basically a Binary matrix with row/col numbers and the number of qubits in the first position

include/cliffordsynthesis/CliffordSynthesizer.hpp
IsFairy: This is something thats in my todos, but something thats a little bit less priority than getting the PR merge ready at the moment.

burgholzer: To be honest, I believe that change is part of the PR being merge-ready. At the moment, the structure in this file is a bit of a mess that's particularly hard to review/grasp.
I guess it's best to resolve the other remarks first, but I really see this as an important part to make all of this seem a little less "hacky".

examples/cliffordexamples/base-tableau.tabl
IsFairy: Ill change the indexes to start from 0, but for the rest, I like that it is more readable that way (also possibly easier to convert into a csv with the separators)

examples/cliffordexamples/destabilizer.txt
IsFairy: Basically just because its the first thing qiskit outputs

include/cliffordsynthesis/CliffordSynthesizer.hpp
IsFairy: I've tried putting the different Targets (Gates, Depth, Fidelity) in different files, but found that it does not improve readability of the CliffordSynthesizer class that much, as it only removes three methods effectively. 
the main complexity still remains (with single and multi gate constraints and the assertion of the tableau)

include/Tableau.hpp
```diff
@@ -0,0 +1,120 @@
+/*
+* This file is part of the MQT QMAP library which is released under the MIT license.
+* See file README.md or go to https://iic.jku.at/eda/research/ibm_qx_mapping/ for more information.
```
burgholzer, line 3: the link in the header still points to a JKU site. It would be great if you could swap it out with the corresponding cda-tum link.

include/configuration/SynthesisConfiguration.hpp
```diff
@@ -0,0 +1,55 @@
...
+        j["chooseBest"] = chooseBest;
+        j["useEmbedding"] = useEmbedding;
+        j["nqubits"] = nqubits;
+        j["initialTimesteps"] = initialTimesteps;
+        j["verbosity"] = verbosity;
+        j["strategy"] = strategy;
+        j["target"] = target;
+        j["method"] = method;
+        j["targetTableau"] = targetTableau.toString();
+        j["initialTableau"] = initialTableau.toString();
```
burgholzer, line 47: from past experience it is better to use *snake_case* for these strings here, since the json dump is frequently exported via the Python bindings and that is the standard naming convention there. Please also check for all the other new structs.

burgholzer (changes requested):
This, again, is a rather incomplete review and includes comments on stuff that I stumbled upon while judging the current structure of the Clifford synthesis library.

I would propose the following structure (demonstrated here for the include files. the same applies to the src directory):
 - include/cliffordsynthesis/
   - Tableau.hpp
   - Configuration.hpp
   - Method.hpp (preferably as "Engine.hpp" or "ReasoningEngine.hpp")
   - Strategy.hpp (preferably as "OptimizationStrategy")
   - Target.hpp (preferably as "TargetMetric.hpp")
   - Results.hpp

All of these should contain structs/classes that have the same name as the file and be placed under a common namespace (to discern them from the mapping equivalents). The namespace could for example be `cs::` in order to stay with the two letter abbreviations (`dd`, `qc`, `ec`,... ).
The namespace and placing everything in a common subfolder safes the annoying `Synthesis` prefix in all names.

As for the interesting part, i.e., the Synthesizer itself: I would definitely combine the `Synthesizer.hpp` and the `CliffordSynthesizer.hpp` files and classes. To be honest, I expected the splitting of the target metrics to have more of an effect. As of now, I does look like that hardly makes sense and the better solution is to combine all of it in the `CliffordSynthesizer.hpp`. Maybe it's possible to eliminate some of the code duplication between strategies and target metrics by factoring out something like "termination criterion", "stopping condition", etc.

include/cliffordsynthesis/CliffordSynthesizer.hpp
IsFairy: The main problem I had with using OpTypes was that it was harder to discern controlled gates and single qubit gates, so this Class was the easiest solution

burgholzer: Hm. I see your point. Not totally happy with the solution, but I also don't have a good alternative in mind. Let's keep this for now.

Just as a side note: this list is still missing the SWAP gate.

IsFairy: SWAP is not a Gate the Synthesizer produces, which is why its not included.

burgholzer:
This is just a checkpoint for my in-depth look through the code. I started pretty much from the high level and began to refactor and fix certain parts as I stumbled upon them. Along the way I probably broke more things than I fixed, but I plan to continue working on the PR during the course of this week.

What I mainly did, was add some documentation of how I would like all the stuff to work at least from the minimal point of view of gate-optimal synthesis. Now I need some time to really dig deep into the code and see where some aspects can be improved.

I try to keep related changes in as small commits as possible so that the changes stay transparent.

burgholzer:
Another checkpoint from my end. There are still lots of places in the code that need improvement.
Over the course of the past days, I discovered several subtle bugs in the implementation that were not covered by tests. Judging from the complexity of some of the functions and the lack of proper tests, I expect more of those in the code.

One conceptional thing I am still debating with myself:
the distinction between `synthesize` and `optimize` is so subtle, that it almost does not make sense and creates more duplication than it helps as a separation of concerns. 
In the very end, the best solution might be to have this distinction between optimization and synthesis _only_ in the end-user facing functions and unify the bindings as well as the C++ functions.

I'll continue working on this over the weekend.

include/cliffordsynthesis/Results.hpp
```diff
@@ -0,0 +1,157 @@
+/*
+* This file is part of the MQT QMAP library which is released under the MIT license.
+* See file README.md or go to https://www.cda.cit.tum.de/research/ibm_qx_mapping/ for more information.
+*/
+#pragma once
```
burgholzer, line 5: Since you use `#pragma once` here in one of the new files: try to stay consistent and use it everywhere instead of header guards.

include/cliffordsynthesis/Results.hpp
```diff
@@ -0,0 +1,157 @@
...
+                        os << ",";
+                    }
+                    os << "\"" << std::to_string(f2) << "\"";
+                    skipfirst2 = false;
+                }
+                os << "]";
+                skipfirst = false;
+            }
+            os << "]" << std::endl;
+            os << "}}" << std::endl;
```
burgholzer, line 121: avoid the redundancy of manually building a JSON entry here. Just write the JSON function properly with the desired output and use the corresponding `dump` function of the JSON object.

src/cliffordsynthesis/TargetMetricHandler.cpp
```diff
@@ -0,0 +1,142 @@
...
+*/
+
+#include "cliffordsynthesis/TargetMetricHandler.hpp"
+
+#include "cliffordsynthesis/GateEncoding.hpp"
+namespace cs {
+
+    void TargetMetricHandler::makeTargetMetric(const SynthesisData& data, const Configuration& configuration) {
+        bool useMaxSat = configuration.strategy == OptimizationStrategy::UseMinimizer || configuration.strategy == OptimizationStrategy::SplitIter;
+        bool onlyCnot  = configuration.target == TargetMetric::TWO_QUBIT_GATES;
```
burgholzer, line 13: these can be const

src/cliffordsynthesis/TargetMetricHandler.cpp
```diff
@@ -0,0 +1,142 @@
...
+        if (useMaxSat) {
+            logicbase::LogicTerm cost = logicbase::LogicTerm(0);
+            for (unsigned int gateStep = 1; gateStep < data.timesteps + 1; ++gateStep) {
+                for (unsigned int a = 0; a < data.nqubits; ++a) {
+                    if (!onlyCNOT) {
+                        for (auto gate: Gates::SINGLE_QUBIT_WITHOUT_NOP) {
+                            cost = cost + data.gS[gateStep][Gates::toIndex(gate)][a];
+                        }
+                    }
+                    for (unsigned int b = 0; b <= a; ++b) {
```
burgholzer, line 40: use `std::size_t` instead of `unsigned int`. Please check throughout this file for similar occurrences

src/cliffordsynthesis/TargetMetricHandler.cpp
```diff
@@ -0,0 +1,142 @@
...
+                break;
+            case TargetMetric::FIDELITY:
+                makeFidelityMetric(data, useMaxSat, configuration.architecture);
+                break;
+        }
+    }
+    void TargetMetricHandler::makeGateMetric(const SynthesisData& data, bool useMaxSat, bool onlyCNOT) {
+        logicbase::LogicTerm changes = logicbase::LogicTerm(true);
+        // COST
+        if (useMaxSat) {
```
burgholzer, line 31: this function does nothing if `useMaxSat == false`. The same holds for the other functions in this file.
Just remove the whole parameter from these functions and make the corresponding check once at the call site.
Safes duplication, unnecessary parameters and a level of indentation.

src/cliffordsynthesis/TargetMetricHandler.cpp
```diff
@@ -0,0 +1,142 @@
...
+                        anyGate = anyGate && !data.gS[gateStep][Gates::toIndex(gate)][a];
+                    }
+                    for (unsigned int b = 0; b <= a; ++b) {
+                        if (a == b) {
+                            continue;
+                        }
+                        anyGate = anyGate && !data.gTwoQubit[gateStep][a][b] && !data.gTwoQubit[gateStep][b][a];
+                    }
+                }
+                cost = cost + LogicTerm::ite(anyGate, LogicTerm(5), LogicTerm(0));
```
burgholzer, line 70: I might be stupid, but why is the cost increased by `5` here if there is a gate at a particular timestep?
Doesn't this just unnecessarily increase the cost?

src/cliffordsynthesis/TargetMetricHandler.cpp
```diff
@@ -0,0 +1,142 @@
...
+        if (!architecture.isArchitectureAvailable() || !architecture.isCalibrationDataAvailable()) {
+            util::fatal("No fidelity architecture specified in coupling map.");
+        }
+        // COST
+        if (useMaxSat) {
+            logicbase::LogicTerm cost = logicbase::LogicTerm(0);
+            // For each edge in the coupling map, get the fidelity cost
+            for (const auto& edge: data.reducedCM) {
+                logicbase::LogicTerm fidelity = logicbase::LogicTerm(
+                        (1 - std::log(architecture.getFidelityTable()[edge.first][edge.second])) * 1000);
```
burgholzer, line 86: The fidelity table already stores the fidelity and not the error rate. So I believe the `1 - ...` is wrong here.
Please also check for the corresponding calls down below.

burgholzer, line 86: The scaling factor `1000` is a magic number and should probably be configurable. Furthermore, is it guaranteed at all that this factor is enough to get a somewhat reasonable cost function if the gate fidelities get high enough?
Say I have a single-qubit gate fidelity of 0.9999 and a two-qubit gate fidelity of 0.999. Then single- and two-qubit gates would be no different under this scaling.

src/cliffordsynthesis/TargetMetricHandler.cpp
```diff
@@ -0,0 +1,142 @@
...
+                auto a = std::find(data.qubitChoice.begin(), data.qubitChoice.end(), edge.first);
+                auto b = std::find(data.qubitChoice.begin(), data.qubitChoice.end(), edge.second);
```
burgholzer, line 88: can be const

src/cliffordsynthesis/TargetMetricHandler.cpp
```diff
@@ -0,0 +1,142 @@
...
+                if (a == data.qubitChoice.end() || b == data.qubitChoice.end()) {
+                    util::fatal("Coupling map contains invalid qubit.");
+                }
+                // at each time t if there is a gate on the edge, add the cost
+                for (unsigned int gateStep = 0; gateStep < data.timesteps; ++gateStep) {
+                    cost = cost + (data.gTwoQubit[gateStep][std::distance(data.qubitChoice.begin(), a)][std::distance(data.qubitChoice.begin(), b)] * fidelity);
```
burgholzer, line 94: precalculate the distances outside of the for loop!

src/cliffordsynthesis/TargetMetricHandler.cpp
```diff
@@ -0,0 +1,142 @@
...
+            cost = logicbase::LogicTerm(0);
+            for (unsigned int gateStep = 0; gateStep < data.timesteps; ++gateStep) {
+                for (unsigned int a = 0; a < data.nqubits; ++a) {
+                    cost = cost + data.gS[gateStep][1][a] + data.gS[gateStep][2][a];
+                    for (unsigned int b = 0; b < data.nqubits; ++b) {
+                        cost = cost + data.gTwoQubit[gateStep][a][b];
+                    }
+                }
+            }
+            dynamic_cast<logicbase::LogicBlockOptimizer*>(data.lb.get())->maximize(cost);
```
burgholzer, line 118: it's nowhere explained nor is it clear what the purpose of this additional optimization is. why is this needed and what does it do? please add some comments to the code.

src/cliffordsynthesis/TargetMetricHandler.cpp
```diff
@@ -0,0 +1,142 @@
...
+                    currentResults = results;
+                }
+                break;
+            case TargetMetric::DEPTH:
+                if ((results.sat && results.depth < currentResults.depth) || currentResults.depth == 0) {
+                    currentResults = results;
+                }
+                break;
+            case TargetMetric::FIDELITY:
+                if ((results.sat && results.fidelity < currentResults.fidelity) || currentResults.fidelity == 0) {
```
burgholzer, line 135: this is the wrong direction in the inequality...

src/cliffordsynthesis/ExactStrategy.cpp
```diff
@@ -0,0 +1,86 @@
...
+            case OptimizationStrategy::MinMax:
+                runBinarySearch(timesteps, reducedCM, qubitChoice, configuration, synthesizer);
+                break;
+            default:
+                throw std::runtime_error("Unknown optimization strategy");
+        }
+    }
+    void ExactStrategy::runMaxSat(std::size_t timesteps, const CouplingMap& reducedCM, const QubitSubset& qubitChoice, const Configuration& configuration, CliffordSynthesizer& synthesizer) {
+        DEBUG() << "Running minimizer" << std::endl;
+        Results r = synthesizer.mainOptimization(timesteps, reducedCM, qubitChoice, configuration.targetTableau, configuration.initialTableau,
```
burgholzer, line 31:
```suggestion
        const auto r = synthesizer.mainOptimization(timesteps, reducedCM, qubitChoice, configuration.targetTableau, configuration.initialTableau,
```

src/cliffordsynthesis/ExactStrategy.cpp
```diff
@@ -0,0 +1,86 @@
...
+                                                 configuration);
+        TargetMetricHandler::updateResults(configuration, r, synthesizer.optimalResults);
+    }
+    void ExactStrategy::runStartLow(std::size_t timesteps, const CouplingMap& reducedCM, const QubitSubset& qubitChoice, const Configuration& configuration, CliffordSynthesizer& synthesizer) {
```
burgholzer, line 35:
```suggestion
    void ExactStrategy::runStartLow(const std::size_t timesteps, const CouplingMap& reducedCM, const QubitSubset& qubitChoice, const Configuration& configuration, CliffordSynthesizer& synthesizer) {
```

please also fix that for the other functions in this file

src/cliffordsynthesis/ExactStrategy.cpp
```diff
@@ -0,0 +1,86 @@
...
+        DEBUG() << "Running start low" << std::endl;
+        Results r;
+        while (r.result != logicbase::Result::SAT || r.result == logicbase::Result::NDEF) {
+            DEBUG() << "Current t=" << timesteps << std::endl;
+            r = synthesizer.mainOptimization(timesteps, reducedCM, qubitChoice, configuration.targetTableau, configuration.initialTableau, configuration);
+            TargetMetricHandler::updateResults(configuration, r, synthesizer.optimalResults);
+            if (r.result == logicbase::Result::UNSAT) {
+                timesteps *= 1.5;
```
burgholzer, line 43: should these factors be configurable? same for the other strategy

src/cliffordsynthesis/ExactStrategy.cpp
```diff
@@ -0,0 +1,86 @@
...
+            } else {
+                timesteps = oldTimesteps;
+            }
+        }
+    }
+    void ExactStrategy::runBinarySearch(std::size_t timesteps, const CouplingMap& reducedCM, const QubitSubset& qubitChoice, const Configuration& configuration, CliffordSynthesizer& synthesizer) {
+        DEBUG() << "Running minmax" << std::endl;
+        Results r;
+        auto    t     = static_cast<int>(timesteps);
+        auto    upper = static_cast<int>(timesteps);
```
burgholzer, line 67: use some fixed width integer type instead of int here

src/cliffordsynthesis/HeuristicStrategy.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include "cliffordsynthesis/HeuristicStrategy.hpp"
+namespace cs {
+    void cs::HeuristicStrategy::runHeuristicStrategy(const CouplingMap& reducedCM, const std::vector<uint16_t>& qubitChoice, const Configuration& configuration, CliffordSynthesizer& synthesizer) {
+        switch (configuration.strategy) {
+            case OptimizationStrategy::SplitIter:
+                runSplitIter(reducedCM, qubitChoice, configuration, synthesizer);
+                break;
+            default:
+                throw std::runtime_error("Unknown optimization strategy");
+        }
```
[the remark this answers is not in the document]
burgholzer: This should be easy to fix

src/cliffordsynthesis/HeuristicStrategy.cpp
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    void HeuristicStrategy::runSplitIter(const CouplingMap& reducedCM, const QubitSubset& qubitChoice, const Configuration& configuration, CliffordSynthesizer& synthesizer) {
```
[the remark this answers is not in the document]
burgholzer: this should also be easy to fix

src/cliffordsynthesis/HeuristicStrategy.cpp
```diff
@@ -0,0 +1,127 @@
...
+            for (size_t i = 0; i * circuitSplit < circuit.getNindividualOps();
+                 i += nThreads) {
+                threads.clear();
+                DEBUG() << "Currently at " << i * circuitSplit << " of "
+                        << circuit.getNindividualOps() << std::endl;
+                for (int j = 0; j < nThreads; j++) {
+                    auto* r = new Results();
+                    auto* t = new std::thread(runSplinter, i, circuitSplit,
+                                              split, std::ref(reducedCM), std::ref(qubitChoice),
+                                              std::ref(circuit), r, &synthesizer, configuration);
```
burgholzer, line 51: use automatic memory management via unique pointers or, if needed, shared pointers

src/cliffordsynthesis/HeuristicStrategy.cpp
```diff
@@ -0,0 +1,127 @@
...
+                if (circuit.getNindividualOps() ==
+                    localResultCircuit.getNindividualOps()) {
+                    break;
+                }
+                circuit = localResultCircuit.clone();
+            }
+        }
+        std::stringstream ss;
+        circuit.dump(ss, qc::Format::OpenQASM);
+        synthesizer.optimalResults.resultStringCircuit = ss.str();
```
burgholzer, line 116: there is a function for that somewhere in the newly added coed

src/cliffordsynthesis/HeuristicStrategy.cpp
```diff
@@ -0,0 +1,127 @@
...
+        synthesizer.optimalResults.resultTableaus.emplace_back(configuration.targetTableau);
+        synthesizer.optimalResults.gateCount = circuit.getNindividualOps();
+        synthesizer.optimalResults.result    = logicbase::Result::SAT;
+    }
+    void HeuristicStrategy::runSplinter(int i, unsigned int circSplit, unsigned int split, const CouplingMap& reducedCM, const QubitSubset& qubitChoice, qc::QuantumComputation& circuit, Results* r, CliffordSynthesizer* opt, const Configuration& configuration) {
+        Tableau targetTableau{circuit, 0, static_cast<std::size_t>((i + 1U)) * circSplit};
+        Tableau initTableau{circuit, 0, static_cast<std::size_t>(i) * circSplit};
```
burgholzer, line 123: these can be const

src/cliffordsynthesis/HeuristicStrategy.cpp
```diff
@@ -0,0 +1,127 @@
...
+            return;
+        }
+        DEBUG() << "Running split iter" << std::endl;
+        Tableau                   fullTableau  = configuration.targetTableau;
+        auto                      circuitSplit = static_cast<unsigned int>(std::log(configuration.targetCircuit.getNindividualOps()));
+        int                       split        = std::min(5, configuration.nqubits / 2);
+        std::vector<std::thread*> threads;
+        std::vector<Results*>     results;
+        int                       nThreads = configuration.nThreads;
+        qc::QuantumComputation    circuit  = configuration.targetCircuit.clone();
```
burgholzer, line 31: is this clone really necessary?

src/cliffordsynthesis/GateEncoding.cpp
```diff
@@ -0,0 +1,364 @@
...
+                if (a == b) {
+                    continue;
+                }
+                changes = changes && (data.x[gateStep][b] == data.x[gateStep - 1][b]);
+                changes = changes && (data.z[gateStep][b] == data.z[gateStep - 1][b]);
+            }
+
+            changes = changes && (data.r[gateStep] == data.r[gateStep - 1]);
+            changes = logicbase::LogicTerm::implies(data.gS[gateStep][0][a], changes);
+            data.lb->assertFormula(changes);
```
burgholzer, line 53: please split this huge function into multiple smaller ones that each have a dedicated purpose.

Also there seems to be lots of code duplication between these individual functions. Try to reduce this as much as possible by encapsulating shared functionality.

As of now, this file is hardly readable and lacks structure. this makes it really hard to check all these constraints for correctness.

src/cliffordsynthesis/GateEncoding.cpp
```diff
@@ -0,0 +1,364 @@
...
+    for (unsigned int gateStep = 1; gateStep < data.timesteps + 1; ++gateStep) {
+        logicbase::LogicTerm rChanges = data.r[gateStep - 1];
+        auto                 aIt      = data.qubitChoice.cbegin();
+        for (unsigned int a = 0; a < data.nqubits; ++a) {
+            // NO GATE
+            changes = logicbase::LogicTerm(true);
+            changes = changes && (data.x[gateStep][a] == data.x[gateStep - 1][a]);
+            changes = changes && (data.z[gateStep][a] == data.z[gateStep - 1][a]);
+            changes = logicbase::LogicTerm::implies(data.gS[gateStep][0][a], changes);
+            data.lb->assertFormula(changes);
```
burgholzer, line 253: these seem rather redundant with the constraints above. lots of this duplication can be saved by extracting the common functionality into functions.

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -0,0 +1,331 @@
...
+        }
+
+        auto                          totalEnd = std::chrono::high_resolution_clock::now();
+        std::chrono::duration<double> diff     = totalEnd - totalStart;
+        INFO() << "Time: " << diff.count() << std::endl;
+        optimalResults.totalSeconds = diff.count();
+    }
+
+    Results CliffordSynthesizer::mainOptimization(
+            std::size_t          timesteps,
```
burgholzer, line 77: const

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -0,0 +1,331 @@
...
+                params.addParam("maxres.pivot_on_correction_set", false);
+                lb = logicutil::getZ3LogicOptimizer(success, true, params);
+            } else {
+                logicutil::Params params;
+                params.addParam("threads", static_cast<unsigned>(configuration.nThreads / 2));
+                lb = logicutil::getZ3LogicBlock(success, true, params);
+            }
+        } else {
+            return Results{};
+        }
```
burgholzer, line 101: if no other reasoning engines are currently supported, why include them at all as a configurable option?

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -0,0 +1,331 @@
...
+        std::stringstream rName{};
+        for (std::size_t k = 0U; k < timesteps + 1U; ++k) {
+            x.emplace_back();
+            z.emplace_back();
+            for (std::size_t i = 0U; i < configuration.nqubits; ++i) {
+                xName.str("");
+                zName.str("");
+                xName << "x_" << k << "_" << i;
+                zName << "z_" << k << "_" << i;
+                x.back().push_back(
```
burgholzer, line 131: why not use `emplace_back` here as well? same question holds for all similar calls down below.

burgholzer (changes requested):
Sorry that there has been such a long period of silence here. I really wanted to work further on this and improve it myself, but I just don't have the time anymore for day-long coding sessions like these.

However, I made it through the pull request with a rather thorough review. You can find the corresponding comments down below. These should provide a solid basis for getting this PR to a merge-ready state.
In summary, I see the following major open points:
 - Internal refactoring of the encoding part to reduce code duplication and encapsulate functionality and make it easier to judge the correctness
 - Proper tests that actually test the expected behavior of the synthesizer on various real-world examples (more in the comments below)
 - More elaborate documentation

src/cliffordsynthesis/ExactStrategy.cpp
IsFairy: this does not work, as the functions do in fact change their local copy of timesteps

src/cliffordsynthesis/ExactStrategy.cpp
IsFairy: this does not work without changing updateResults, which does not always get a const

src/cliffordsynthesis/ExactStrategy.cpp
burgholzer: Alright, then this is fine.

src/cliffordsynthesis/ExactStrategy.cpp
burgholzer: Any argument against changing the updateResults signature?

IsFairy: Not at all, I'll do that

src/cliffordsynthesis/ExactStrategy.cpp
IsFairy: Seems like a good idea, will be implemented

src/cliffordsynthesis/HeuristicStrategy.cpp
IsFairy: Fixed

src/cliffordsynthesis/HeuristicStrategy.cpp
IsFairy: Fixed

src/cliffordsynthesis/HeuristicStrategy.cpp
IsFairy: Fixed

test/test_cliffordsynthesis.cpp
```diff
@@ -0,0 +1,325 @@
...
+        londonOptimizer = std::make_unique<CliffordSynthesizer>();
+
+        qx4Optimizer = std::make_unique<CliffordSynthesizer>();
+    }
+};
+
+INSTANTIATE_TEST_SUITE_P(
+        CliffordSynthesizer, TestCliffordSynthesis,
+        testing::Values(
+                "stabilizer.txt"));
```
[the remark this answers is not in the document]
IsFairy: changed the test signature to make it easier to debug as well

src/cliffordsynthesis/TargetMetricHandler.cpp
```diff
@@ -27,22 +27,20 @@ namespace cs {
...
-                    }
+                for (auto gate: Gates::SINGLE_QUBIT_WITHOUT_NOP) {
+                    cost = cost + data.gS[gateStep][Gates::toIndex(gate)][a];
                 }
                 for (std::size_t b = 0; b <= a; ++b) {
                     if (a == b) {
                         continue;
                     }
-                    cost = cost + data.gTwoQubit[gateStep][a][b] + data.gTwoQubit[gateStep][b][a];
+                    cost = cost + (data.gTwoQubit[gateStep][a][b] + data.gTwoQubit[gateStep][b][a]) * logicbase::LogicTerm(onlyCNOT ? cnotMultiplier : 1);
```
burgholzer, line 43: This change does not make sense to me. If I only care for the CNOT count, then it does not matter how many single-qubit gates there are.
With this change, a circuit with `x+1` CNOT gates and `y-11` single-qubit gates would be preferred over a circuit with `x` CNOT gates and `y` single-qubit gates.
Thus, yielding a sub-optimal solution.

Please revert this, if there is no argument against the above analysis.

burgholzer:
Please resolve the conflicts in this PR so that CI is triggered properly.

burgholzer (changes requested):
I just made it through a quick round of review for the latest roughly 30 commits.
You can find detailed comments below. It would be really great to get this done soon. I have looked at the code for so many times now...

test/test_cliffordsynthesis.cpp
```diff
@@ -0,0 +1,325 @@
...
+    std::string testCalibrationDir  = "./calibration/";
+    std::string testExampleDir      = "./examples/cliffordexamples/";
+
+    Architecture ibmqYorktown{};
+    Architecture ibmqLondon{};
+    Architecture ibmQX4{};
+
+    std::unique_ptr<CliffordSynthesizer> qx4Optimizer{};
+    std::unique_ptr<CliffordSynthesizer> yorktownOptimizer{};
+    std::unique_ptr<CliffordSynthesizer> londonOptimizer{};
```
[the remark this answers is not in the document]
burgholzer: To be clear: what I really meant here, was a function that allows to construct an architecture with, let's say, `0.99` two-qubit fidelity, `0.999` single-qubit fidelity, and `0.995` readout fidelity.

src/cliffordsynthesis/TargetMetricHandler.cpp
```diff
@@ -0,0 +1,130 @@
...
+                for (std::size_t b = 0; b <= a; ++b) {
+                    if (a == b) {
+                        continue;
+                    }
+                    anyGate = anyGate && !data.gTwoQubit[gateStep][a][b] && !data.gTwoQubit[gateStep][b][a];
+                }
+            }
+            cost = cost + LogicTerm::ite(anyGate, LogicTerm(1), LogicTerm(0));
+        }
+        dynamic_cast<LogicBlockOptimizer*>(data.lb.get())->maximize(cost);
```
burgholzer, line 71: why the h... is there a `maximize` call here?!

src/cliffordsynthesis/TargetMetricHandler.cpp
```diff
@@ -0,0 +1,130 @@
...
+            logicbase::LogicTerm fidelity =
+                    logicbase::LogicTerm((std::log(architecture.getSingleQubitFidelities()[a])) * fidelityScaling);
+            // at each time t if there is a gate on a, add the cost
+            for (std::size_t gateStep = 0; gateStep < data.timesteps; ++gateStep) {
+                for (auto gate: Gates::SINGLE_QUBIT_WITHOUT_NOP) {
+                    cost = cost + (data.gS[gateStep][Gates::toIndex(gate)][a] * fidelity);
+                }
+            }
+        }
+        dynamic_cast<logicbase::LogicBlockOptimizer*>(data.lb.get())->minimize(cost);
```
burgholzer, line 107: and why the h... is there a `minimize` call here?!

burgholzer:
Just merged in the newest changes from `main` and fixed a couple of bugs.
I am starting to factor out some changes from this PR into smaller PRs that can already be incorporated into `main` while the rest of the code is still being fixed.

The `mypy` job is currently failing since stubs for the synthesis functionality in the python bindings are missing from the newly introduced stubs file.

src/cliffordsynthesis/TargetMetricHandler.cpp
burgholzer: Resolved in the latest series of commits.

src/cliffordsynthesis/TargetMetricHandler.cpp
burgholzer: Resolved in the latest series of commits.

src/utils.cpp
```diff
@@ -213,3 +213,23 @@ std::string escapeChars(const std::string& s, const std::string& chars) {
     }
     return ss.str();
 }
+std::uint16_t getSingleQubitOperations(const qc::QuantumComputation& qc) {
```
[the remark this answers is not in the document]
burgholzer: Added in https://github.com/cda-tum/qfr/pull/220

burgholzer:
Superseded by #196.

PR #79, opened 2022-06-29
✨ Integrate LogicBlocks

IsFairy:
This PR adds the https://github.com/IsFairy/LogicBlocks submodule as a drop-in replacement for the existing Z3 code.
It effectively replaces all existing Z3 code with a corresponding wrapper that allows for higher levels of abstraction and (in the future) for employing different solvers in QMAP.


include/exact/ExactMapper.hpp
```diff
@@ -6,7 +6,8 @@
 #ifndef EXACT_MAPPER_hpp
 #define EXACT_MAPPER_hpp
 
-#include "Encodings.hpp"
+#include "Encodings/Encodings.hpp"
+#include "LogicBlock/LogicBlock.hpp"
```
burgholzer, line 10: These includes can be moved to the source file as they are not needed in the header itself.

include/exact/ExactMapper.hpp
```diff
@@ -18,6 +19,10 @@
 #include <unordered_set>
 #include <z3++.h>
```
burgholzer, line 20: this include can be removed
```suggestion
```

include/exact/ExactMapper.hpp
```diff
@@ -18,6 +19,10 @@
...
 
+#ifdef Z3_FOUND
+    #include "LogicBlock/Z3Logic.hpp"
+#endif
+
```
burgholzer, line 25: This include can also be moved to the source file as it is only needed there.

include/exact/ExactMapper.hpp
```diff
@@ -18,6 +19,10 @@
...
 using namespace z3;
 using matrix      = std::vector<expr_vector>;
```
burgholzer, line 27: if z3 is not available, this does not make sense. so this should also be part of an `#ifdef Z3_FOUND` block. if it is needed at all anymore.
in order to comply with the changes above, this `using` directive should also be moved to the corresponding source file.

src/Architecture.cpp
```diff
@@ -97,7 +97,7 @@ void Architecture::loadProperties(const std::string& filename) {
 }
 
 void Architecture::loadProperties(std::istream&& is) {
-    static auto SingleQubitGates = {"id", "u1", "u2", "u3", "rz", "sx", "x"};
+    static const std::vector<std::string> SingleQubitGates = {"id", "u1", "u2", "u3", "rz", "sx", "x"};
```
burgholzer, line 100: any particular reason for this change? I mean, the addition of `const` totally makes sense, but why drop the `auto`?

src/Architecture.cpp
```diff
@@ -121,7 +121,7 @@ void Architecture::loadProperties(std::istream&& is) {
         properties.qubitFrequency.set(qubitNumber, std::stod(data[3]));
         properties.readoutErrorRate.set(qubitNumber, std::stod(data[4]));
         // csv file reports average single qubit fidelities
-        for (const auto& operation: SingleQubitGates) {
+        for (const std::string& operation: SingleQubitGates) {
```
burgholzer, line 124: Similar question as above: why the change here?

src/exact/ExactMapper.cpp
```diff
@@ -327,13 +327,28 @@ void ExactMapper::map(const Configuration& settings) {
...
+    p.set("maxres.hill_climb", true);
+    p.set("maxres.pivot_on_correction_set", false);
+    opt.set(p);
+    lb = new Z3LogicOptimizer(c, opt, true);
+#endif
 
     std::vector<unsigned short>                        pi(qubitChoice.begin(), qubitChoice.end());
     unsigned long long                                 piCount{};
-    unsigned long long                                 internalPiCount{};
+    unsigned long long                                 internalPiCount;
```
burgholzer, line 351: any particular reason for this change? it somehow breaks consistency with how all other values are initialized.

src/exact/ExactMapper.cpp
```diff
@@ -389,115 +405,105 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
...
-    opt.set(p);
-
     //////////////////////////////////////////
     /// 	Consistency Constraints			//
     //////////////////////////////////////////
     if (config.encoding == Encoding::Naive) {
         for (unsigned long k = 0; k < reducedLayerIndices.size(); ++k) {
             for (unsigned long i = 0; i < qubitChoice.size(); ++i) {
-                expr rowConsistency = c.int_val(0);
+                LogicTerm rowConsistency = LogicTerm(0);
```
burgholzer, line 429:
```suggestion
                auto rowConsistency = LogicTerm(0);
```

src/exact/ExactMapper.cpp
```diff
@@ -327,13 +327,28 @@ void ExactMapper::map(const Configuration& settings) {
 
 void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice, const CouplingMap& rcm, MappingResults& choiceResults, std::vector<std::vector<std::pair<unsigned short, unsigned short>>>& swaps, long unsigned int limit, unsigned int timeout) {
     const auto& config = results.config;
-
-    // Z3 context
-    context c;
+    using namespace logicbase;
+    // LogicBlock
+    LogicBlock* lb;
```
burgholzer, line 332:
```suggestion
    std::unique_ptr<LogicBlock> lb;
```
Use C++ facilities to automatically manage memory. This requires some changes in the LogicBlocks repository (specifically the `Encodings` files).
Instead of passing `LogicBlock*` either pass `std::unique_ptr<LogicBlock>&` or `LogicBlock&` and adapt the call sites in QMAP correspondingly.

src/exact/ExactMapper.cpp
```diff
@@ -327,13 +327,28 @@ void ExactMapper::map(const Configuration& settings) {
...
+    z3::optimize opt(c);
+    z3::params   p(c);
+    LogicTerm::termType = TermType::BASE;
+    p.set("timeout", timeout);
+    p.set("pb.compile_equality", true);
+    p.set("pp.wcnf", true);
+    p.set("maxres.hill_climb", true);
+    p.set("maxres.pivot_on_correction_set", false);
+    opt.set(p);
+    lb = new Z3LogicOptimizer(c, opt, true);
```
burgholzer, line 346:
```suggestion
    lb = std::make_unique<Z3LogicOptimizer>(c, opt, true);
```
see above

src/exact/ExactMapper.cpp
burgholzer, line 332: if Z3 is not found, the code is not functional at the moment, right?
Shouldn't this be avoided in order to loosen the dependency on Z3?

src/exact/ExactMapper.cpp
```diff
@@ -389,115 +405,105 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
...
-                                     ite(x[k][i][j], c.int_val(1), c.int_val(0));
+                                     LogicTerm::ite(x[k][i][j], LogicTerm(1), LogicTerm(0));
                 }
-                opt.add(rowConsistency.simplify() <= 1);
+                lb->assertFormula(rowConsistency <= LogicTerm(1));
             }
 
             for (unsigned short j = 0; j < qc.getNqubits(); ++j) {
-                expr colConsistency = c.int_val(0);
+                LogicTerm colConsistency = LogicTerm(0);
```
burgholzer, line 438:
```suggestion
                auto colConsistency = LogicTerm(0);
```

src/exact/ExactMapper.cpp
```diff
@@ -360,7 +375,7 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
     /*
 	 auxilary variable declarations
 	*/
-    expr_vector auxvars(c);
+    std::vector<LogicTerm> auxvars;
```
burgholzer, line 378: this is no longer needed as it seems.

src/exact/ExactMapper.cpp
```diff
@@ -507,12 +513,12 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
     ///		Coupling Constraints			//
     //////////////////////////////////////////
     for (unsigned long k = 0; k < reducedLayerIndices.size(); ++k) {
-        expr allCouplings = c.bool_val(true);
+        LogicTerm allCouplings = LogicTerm(true);
```
burgholzer, line 516:
```suggestion
        auto allCouplings = LogicTerm(true);
```

src/exact/ExactMapper.cpp
```diff
@@ -507,12 +513,12 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
...
         for (const auto& gate: layers.at(reducedLayerIndices.at(k))) {
             if (gate.singleQubit())
                 continue;
 
-            expr coupling = c.bool_val(false);
+            LogicTerm coupling = LogicTerm(false);
```
burgholzer, line 521:
```suggestion
            auto coupling = LogicTerm(false);
```

src/exact/ExactMapper.cpp
```diff
@@ -544,15 +550,15 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
         auto& j         = x[k];
         do {
             if (skipped_pi.count(piCount) == 0 || !config.enableSwapLimits) {
-                expr equal = c.bool_val(true);
+                LogicTerm equal = LogicTerm(true);
```
burgholzer, line 553:
```suggestion
                auto equal = LogicTerm(true);
```

src/exact/ExactMapper.cpp
```diff
@@ -562,21 +568,22 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
     // Allow only 1 y_k_pi to be true
     if (config.encoding == Encoding::Naive) {
         for (unsigned long k = 1; k < reducedLayerIndices.size(); ++k) {
-            expr onlyOne    = c.int_val(0);
-            piCount         = 0;
-            internalPiCount = 0;
+            LogicTerm onlyOne = LogicTerm(0);
```
burgholzer, line 571:
```suggestion
            auto onlyOne = LogicTerm(0);
```

src/exact/ExactMapper.cpp
```diff
@@ -604,6 +611,7 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
     piCount         = 0;
     internalPiCount = 0;
     std::vector<std::set<WeightedVar>> weightedVars(reducedLayerIndices.size());
+    LogicTerm                          cost = LogicTerm(0);
```
burgholzer, line 614:
```suggestion
    auto                          cost = LogicTerm(0);
```

src/exact/ExactMapper.cpp
```diff
@@ -613,18 +621,18 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
                 picost *= GATES_OF_UNIDIRECTIONAL_SWAP;
             }
             for (unsigned long k = 1; k < reducedLayerIndices.size(); ++k) {
-                opt.add(!y[k - 1][internalPiCount], picost);
+                cost = cost + LogicTerm::ite(y[k - 1][internalPiCount], LogicTerm(static_cast<int>(picost)), LogicTerm(0));
                 if (config.useBDD)
-                    weightedVars[k].insert(WeightedVar(internalPiCount, static_cast<int>(picost)));
+                    weightedVars[k].insert(WeightedVar(y[k - 1][internalPiCount], static_cast<int>(picost)));
```
burgholzer, line 626:
```suggestion
                    weightedVars[k].emplace(WeightedVar(y[k - 1][internalPiCount], static_cast<int>(picost)));
```

src/exact/ExactMapper.cpp
```diff
@@ -635,13 +643,13 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
                 if (gate.singleQubit())
                     continue;
 
-                expr reverse = c.bool_val(true);
+                LogicTerm reverse = LogicTerm(true);
```
burgholzer, line 646:
```suggestion
                auto reverse = LogicTerm(true);
```

src/exact/ExactMapper.cpp
```diff
@@ -635,13 +643,13 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
...
                 for (const auto& edge: rcm) {
                     auto indexFT = x[k][physicalQubitIndex[edge.first]][gate.target];
                     auto indexSC = x[k][physicalQubitIndex[edge.second]][gate.control];
                     reverse      = reverse && (!indexFT || !indexSC);
                 }
-                opt.add(reverse.simplify(), GATES_OF_DIRECTION_REVERSE);
+                dynamic_cast<LogicBlockOptimizer*>(lb)->weightedTerm(reverse, GATES_OF_DIRECTION_REVERSE);
```
burgholzer, line 652: Is is guaranteed that this dynamic cast is successful? Shouldn't there be a check? otherwise `lb` could just straight-up be a `LogicBlockOptimizer`. Then there is no need for a cast at all.

burgholzer, line 652:
```suggestion
                dynamic_cast<LogicBlockOptimizer*>(lb.get())->weightedTerm(reverse, GATES_OF_DIRECTION_REVERSE);
```
just a thing to not overlook when changing to automated memory management.

src/exact/ExactMapper.cpp
```diff
@@ -662,8 +670,10 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
     //////////////////////////////////////////
     /// 	Solving							//
     //////////////////////////////////////////
-    if (sat == opt.check()) {
-        model m               = opt.get_model();
+    lb->produceInstance();
+    Result res = lb->solve();
+    if (Result::SAT == res) {
+        Model* m              = lb->getModel();
```
burgholzer, line 676:
```suggestion
        const auto m              = lb->getModel();
```
It should be possible to make the generated model `const`. However, this requires some functions in `LogicBlocks` to also be `const`. I currently lack the in-depth knowledge of the `LogicBlocks` library to find out whether that is feasible.

src/exact/ExactMapper.cpp
```diff
@@ -662,8 +670,10 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
     //////////////////////////////////////////
     /// 	Solving							//
     //////////////////////////////////////////
-    if (sat == opt.check()) {
-        model m               = opt.get_model();
+    lb->produceInstance();
+    Result res = lb->solve();
```
burgholzer, line 674:
```suggestion
    const auto res = lb->solve();
```

burgholzer (changes requested):
Thanks for this contribution!
I just quickly went through all the changes and have some questions/remarks/requests. You can find all of them below. Most of them should be an easy fix. Some might require small refactorings in the `LogicBlocks` library.
Let me know if I can help.

src/Architecture.cpp
IsFairy: The auto made problems with some types of compilers, because it was converted to a vector of char* and this in turn only looked at the first character when doing a conversion to a gate. 
For gcc its fine, but msvc complained for some versions of the compiler. So I thought it best to be more clear which type is wanted.

CMakeLists.txt
```diff
@@ -35,26 +35,29 @@ if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
 # search for Z3
 find_package(Z3)
-if(Z3_FOUND AND NOT TARGET z3::z3lib)
+if (Z3_FOUND AND NOT TARGET z3::z3lib)
 	message(STATUS "Found Z3 with version ${Z3_VERSION_STRING}")
 	add_library(z3::z3lib IMPORTED INTERFACE)
 	set_property(TARGET z3::z3lib PROPERTY INTERFACE_INCLUDE_DIRECTORIES ${Z3_CXX_INCLUDE_DIRS})
 	set_property(TARGET z3::z3lib PROPERTY INTERFACE_LINK_LIBRARIES ${Z3_LIBRARIES})
-else()
+	add_definitions(-DZ3_FOUND)
```
burgholzer, line 57: `add_definitions` has been superseded. See https://cmake.org/cmake/help/latest/command/add_compile_definitions.html#command:add_compile_definitions

CMakeLists.txt
```diff
@@ -35,26 +35,29 @@ if (NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
+else ()
 	message(WARNING "Did not find Z3. Exact library and other depending target will not be available")
-endif()
+endif ()
```
burgholzer, line 60: why is the discovery code for Z3 still here if this PR adopts the LogicBlocks module as a replacement? this should most likely be removed or replaced by some checks when/after the LogicBlocks submodule is loaded.

cmake/FindZ3.cmake
```diff
@@ -67,15 +67,15 @@ if (NOT Z3_FOUND)
...
 
-	if(NOT Z3_VERSION_STRING)
+	if (NOT Z3_VERSION_STRING)
 		set(Z3_VERSION_STRING "0.0.0")
-	endif()
+	endif ()
 
-	include (FindPackageHandleStandardArgs)
+	include(FindPackageHandleStandardArgs)
 	find_package_handle_standard_args(Z3
```
burgholzer, line 79: why the whitespace changes throughout this file?

But, more importantly, this file should not be needed at all anymore, since this is all supposed to be handled by the LogicBlocks module.

src/Architecture.cpp
```diff
@@ -121,7 +121,7 @@ void Architecture::loadProperties(std::istream&& is) {
         properties.qubitFrequency.set(qubitNumber, std::stod(data[3]));
         properties.readoutErrorRate.set(qubitNumber, std::stod(data[4]));
         // csv file reports average single qubit fidelities
-        for (const auto& operation: SingleQubitGates) {
+        for (const auto operation: SingleQubitGates) {
```
burgholzer, line 124: why remove the reference here and incur a copy?

src/Architecture.cpp
```diff
@@ -302,6 +302,10 @@ void Architecture::minimumNumberOfSwaps(std::vector<unsigned short>& permutation
         qubits.insert(q);
     }
 
+    if (qubits.size() != permutation.size()) {
+        throw std::runtime_error("Architecture::minimumNumberOfSwaps: permutation contains duplicates");
+    }
```
burgholzer, line 307: this change is not really related to this PR...

src/CMakeLists.txt
```diff
@@ -1,85 +1,85 @@
...
+	            ${CMAKE_CURRENT_SOURCE_DIR}/Architecture.cpp
+	            ${CMAKE_CURRENT_SOURCE_DIR}/Mapper.cpp
+	            ${CMAKE_CURRENT_SOURCE_DIR}/${LIBNAME}/${SRCFILE}.cpp
+	            ${${PROJECT_NAME}_SOURCE_DIR}/include/utils.hpp
+	            ${${PROJECT_NAME}_SOURCE_DIR}/include/Architecture.hpp
+	            ${${PROJECT_NAME}_SOURCE_DIR}/include/Mapper.hpp
+	            ${${PROJECT_NAME}_SOURCE_DIR}/include/MappingResults.hpp
+	            ${${PROJECT_NAME}_SOURCE_DIR}/include/configuration
+	            ${${PROJECT_NAME}_SOURCE_DIR}/include/${LIBNAME}/${SRCFILE}.hpp
+	            ${${PROJECT_NAME}_SOURCE_DIR}/include/heuristic/unique_priority_queue.hpp)
```
burgholzer, line 19: why the whitespace changes throughout this file?

src/CMakeLists.txt
```diff
@@ -1,85 +1,85 @@
...
-            ${CMAKE_CURRENT_SOURCE_DIR}/Encodings.cpp
-            ${${PROJECT_NAME}_SOURCE_DIR}/include/Encodings.hpp)
-    # add Z3 library
-    target_link_libraries(${PROJECT_NAME}_exact_lib PUBLIC z3::z3lib)
+	# exact mapper project library
+	add_qmap_library(exact ExactMapper)
+	# add Z3 library
+	target_link_libraries(${PROJECT_NAME}_exact_lib PUBLIC z3::z3lib)
+	# add LogicBlocks library
+	target_link_libraries(${PROJECT_NAME}_exact_lib PUBLIC logicblocks::Logic logicblocks::Encodings)
```
burgholzer, line 80: It should no longer be necessary to link to Z3 here and to have the inclusion of the Exact Mapper conditioned on `Z3_FOUND` (which shouldn't exist as by the comment above).
It should be enough to link to the LogicBlocks module here and the LogicBlocks library should handle the availability of Z3 and either fail if it is not available or provide a fallback.

test/test_architecture.cpp
```diff
@@ -150,3 +150,10 @@ TEST(TestArchitecture, FullyConnectedTest) {
 
     ASSERT_TRUE(cm.size() == 3 * 2);
 }
+
+TEST(TestArchitecture, MinimumNumberOfSwapsError) {
+    Architecture                                           architecture{};
+    std::vector<unsigned short>                            permutation{1, 1, 2, 3, 4};
+    std::vector<std::pair<unsigned short, unsigned short>> swaps{};
+    EXPECT_THROW(architecture.minimumNumberOfSwaps(permutation, swaps), std::runtime_error);
+}
```
burgholzer, line 159: this change does not seem to be related to this PR.

test/test_general.cpp
```diff
@@ -50,3 +50,13 @@ TEST(General, TestLineParsing) {
...
 }
+
+TEST(General, TestUtilFunctions) {
+    std::vector<unsigned short> data{3, 4, 2, 1};
+    EXPECT_EQ(printPi(data), "(3,4,2,1)");
+
+    std::set<unsigned short> data_set1{3, 4, 2, 1};
+    std::set<unsigned short> data_set2{1, 2, 3};
+    EXPECT_EQ(idx(10, 4, 2, data_set1, data_set2), 120 + 6 + 4);
+    EXPECT_EQ(idx(10, 4, 2, data_set1, 4), 160 + 8 + 6);
```
burgholzer, line 61: While I always appreciate additional tests, these tests
1) are not related to the PR
2) contain lots of magic numbers and it is not quite clear what is being tested here. Less magic numbers and/or some comments would help here

src/exact/ExactMapper.cpp
```diff
@@ -5,6 +5,12 @@
 
 #include "exact/ExactMapper.hpp"
 
+#include "Encodings/Encodings.hpp"
+#include "LogicBlock/LogicBlock.hpp"
```
burgholzer, line 9: This is more of a general remark. From these includes it looks like `Encoding` and `LogicBlock` are part of the QMAP project. However, they originate from the Logicblocks repo. Wouldn't it make sense to place the Logicblocks code in a separate `lb` (or similar) directory similar to how this is done for the DD Package in the QFR for example?

src/exact/ExactMapper.cpp
```diff
@@ -5,6 +5,12 @@
...
+#ifdef Z3_FOUND
```
burgholzer, line 10: As stated earlier, this compile definition should be set by the LogicBlocks module whenever Z3 is available.

src/exact/ExactMapper.cpp
burgholzer: this is still to be addressed

src/exact/ExactMapper.cpp
```diff
@@ -369,16 +385,17 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
...
     for (unsigned long k = 0; k < reducedLayerIndices.size(); ++k) {
         x.emplace_back();
         for (unsigned short Q: qubitChoice) {
-            x.back().emplace_back(c);
+            x.back().emplace_back();
             for (unsigned short q = 0; q < qc.getNqubits(); ++q) {
                 x_name.str("");
                 x_name << "x_" << k << '_' << Q << '_' << q;
-                x.back().back().push_back(c.bool_const(x_name.str().c_str()));
+                x.back().back().push_back(
```
burgholzer, line 397:
```suggestion
                x.back().back().emplace_back(
```

src/exact/ExactMapper.cpp
```diff
@@ -389,115 +406,105 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
...
+    std::stringstream y_name{};
     for (unsigned long k = 1; k < reducedLayerIndices.size(); ++k) {
         y.emplace_back();
         piCount = 0;
         do {
             if (skipped_pi.count(piCount) == 0 || !config.enableSwapLimits) {
                 y_name.str("");
                 y_name << "y_" << k << '_' << piCount;
-                y.back().push_back(c.bool_const(y_name.str().c_str()));
+                y.back().push_back(lb->makeVariable(y_name.str(), CType::BOOL));
```
burgholzer, line 418:
```suggestion
                y.back().emplace_back(lb->makeVariable(y_name.str(), CType::BOOL));
```

src/exact/ExactMapper.cpp
```diff
@@ -544,15 +551,15 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
         auto& j         = x[k];
         do {
             if (skipped_pi.count(piCount) == 0 || !config.enableSwapLimits) {
-                expr equal = c.bool_val(true);
+                LogicTerm equal = LogicTerm(true);
```
burgholzer, line 554:
```suggestion
                auto equal = LogicTerm(true);
```

src/exact/ExactMapper.cpp
```diff
@@ -562,21 +569,22 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
...
                 if (skipped_pi.count(piCount) == 0 || !config.enableSwapLimits) {
-                    onlyOne = onlyOne + ite(y[k - 1][internalPiCount], c.int_val(1), c.int_val(0));
+                    onlyOne = onlyOne + LogicTerm::ite(y[k - 1][internalPiCount], LogicTerm(1), LogicTerm(0));
                     ++internalPiCount;
                 }
                 ++piCount;
             } while (std::next_permutation(pi.begin(), pi.end()));
-            opt.add(onlyOne.simplify() == 1);
+            auto t = onlyOne == LogicTerm(1);
+            lb->assertFormula(t);
```
burgholzer, line 583: why the temporary variable here?
```suggestion
            lb->assertFormula(onlyOne == LogicTerm(1));
```

src/exact/ExactMapper.cpp
```diff
@@ -635,13 +644,13 @@ void ExactMapper::coreMappingRoutine(const std::set<unsigned short>& qubitChoice
...
 
-                expr reverse = c.bool_val(true);
+                auto reverse = LogicTerm(true);
                 for (const auto& edge: rcm) {
                     auto indexFT = x[k][physicalQubitIndex[edge.first]][gate.target];
                     auto indexSC = x[k][physicalQubitIndex[edge.second]][gate.control];
                     reverse      = reverse && (!indexFT || !indexSC);
                 }
-                opt.add(reverse.simplify(), GATES_OF_DIRECTION_REVERSE);
+                lb->weightedTerm(reverse, GATES_OF_DIRECTION_REVERSE);
```
burgholzer, line 653: why is this handled differently than the `cost` above?

burgholzer (changes requested):
Just went through all the changes here and left my comments below with some things to be addressed before this can be merged. 
I still believe we should get this PR in before #78.

One thing I could not annotate in the changed files: The LGTM C++ run is failing for the same reason the CodeQL run was failing: a more recent version of Z3 is needed to build the project and the standard ubuntu repo only provides `4.8.7`.
Two comments on that end:
 - for one, it should not be possible to reach that point of failure at all. The `LogicBlocks` submodule should specify a minimum required version of Z3 so that compilation is not even started when no suitable version is installed.
 - the fix for the LGTM run is similar to your fix for the CodeQL run: `z3` has to be installed from pip and the respective paths have to be set as environment variables in https://github.com/cda-tum/qmap/blob/main/.lgtm.yml

Last but not least, I was wondering whether the "fallback" when z3 is not available should also be tested. Is this sufficiently implemented in the LogicBlocks module?

src/Architecture.cpp
IsFairy: It provides a more helpful error message when something with the encodings goes wrong, either the naive ones or the ones provided by logicblocks. It is thus tangentially related.

test/test_architecture.cpp
IsFairy: See above

IsFairy:
> Last but not least, I was wondering whether the "fallback" when z3 is not available should also be tested. Is this sufficiently implemented in the LogicBlocks module?

The replacement, or fallback for z3 is still in the works for now and not really tested, so for the moment z3 is still required when using the exact mapper. The planned fallback (SMTLibv2 Export) also wont produce a model, so its something that the exact mapper will need to handle anyway.

burgholzer:
> > Last but not least, I was wondering whether the "fallback" when z3 is not available should also be tested. Is this sufficiently implemented in the LogicBlocks module?
> 
> The replacement, or fallback for z3 is still in the works for now and not really tested, so for the moment z3 is still required when using the exact mapper. The planned fallback (SMTLibv2 Export) also wont produce a model, so its something that the exact mapper will need to handle anyway.

I see, makes sense. From that point of view, Z3 will be required for the foreseeable future when using QMAP.
At least until we figure out how to bundle another SMT solver (that works under all operating systems) with the distributed package and automatically employ it instead of Z3.

burgholzer (changes requested):
Two very small requests (in the comments), otherwise this seems almost ready.

LGTM shows two alerts in the LogicBlocks submodule that should be very quick to fix (see https://lgtm.com/projects/g/cda-tum/qmap/snapshot/72b875f873c787a1b50dd6a177f689ac21022440/files/extern/LogicBlocks/?sort=name&dir=ASC&mode=heatmap).

The wheel builds fails because the Z3 library is not picked up. This is most likely due to you deleting the `FindZ3.cmake` file. This leads to the libraries not being found in https://github.com/IsFairy/qmap/blob/974ae4a867603163abd0e901088b981e5b30a6ba/CMakeLists.txt#L51
And, potentially, `Z3_FOUND` not being truthy in https://github.com/IsFairy/qmap/blob/LogicBlocks/mqt/qmap/CMakeLists.txt.
But this is just a guess.

burgholzer (commented):
Overall all the present changes LGTM. I would love to see the coverage check turn green, which probably requires covering 1-2 more lines.
To that end, I was wondering whether the following lines can simply be removed as they do not seem to be used anymore:
https://github.com/IsFairy/qmap/blob/1bf45637a875228c7402a664e3b87fa6bed7eabf/src/utils.cpp#L116-L144

What I was also thinking about: does the https://github.com/IsFairy/qmap/blob/LogicBlocks/include/configuration/Encoding.hpp file and code related to it still fit into this project or should it also be moved to the LogicBlocks submodule?
The same goes for https://github.com/IsFairy/qmap/blob/LogicBlocks/include/configuration/CommanderGrouping.hpp
Most probably, this should be done in a separate PR since it also requires tuning the bindings code here.

burgholzer (approved):

PR #120, opened 2022-09-15
⚗️ Test C++ Coverage

burgholzer:
This is just an experiment whether setting the dedicated C++ test targets for the coverage run improves coverage statistics.


burgholzer:
Coverage is not affected. This is most likely due to `-DBINDINGS=ON` not being passed to the coverage job CMake configuration.

PR #165, opened 2022-10-28
🐛 Fix Qiskit compatibility

burgholzer:
## Description

This PR fixes an issue caused by a breaking change in the newest `qiskit-terra` update. The `_layout` member of `QuantumCircuit`s has been changed to a data class. This caused our circuit import and also the export to fail on the newer version.
This has not been caught earlier because QCEC (which is a test dependency) still capped qiskit-terra at `<0.22`.

In addition to the fix above, this PR loosens many of the dependency versions in order to avoid such situations in the future.

## Checklist:


burgholzer:
In the future, we might want to introduce an additional CI run that tests the lowest qiskit-terra version supported. We could then upload that coverage to also cover the lines missed here in this PR.

PR #204, opened 2022-12-17
✨Heuristic mapper fidelity preparation

EliasLF:
## Description

- isolated the heuristic to HeuristicMapper::Node::updateHeuristicCost for less redundant code and easier expansion with further heuristic types (e.g. the planned fidelity aware approach)
- removed redundant field HeuristicMapper::Node::totalCost (redundant to HeuristicMapper::Node::costFixed)
- added consider_fidelity field to configuration classes

## Checklist:

- [x] The pull request only contains commits that are related to it.
- [x] I have added appropriate tests and documentation.
- [x] I have made sure that all CI jobs on GitHub pass.
- [x] The pull request introduces no new warnings and follows the project's style guidelines.

PR #218, opened 2023-01-20
Fidelity heuristic

EliasLF:
## Description

some major changes in the heuristic mapper:
- adds fidelity-aware mapping (which can be used by loading calibration data for the respective architecture and setting `configuration.considerFidelity = true`) However, for now it is not yet compatible with dynamic initial layout strategies and implements no lookahead.
- in the calculation of cost and heuristic of A*-nodes iteration over all gates of the current layer for each node were replaced by iterating over a pre-computed gate multiplicity vector (which was necessary for the fidelity-aware approach, but also increases performance of the current mapping implementation)
- the constant edge weights in the Dijkstra implementation were replaced with an edge weight matrix (which was necessary for fidelity-aware mapping, but, too, improves the current non-fidelity-aware heuristic implementation as Dijkstra did not differentiate between the cost of forward- and back-edges there until now)

furthermore:
- eliminates (potential) bug previously introduced in #204 where heuristic cost calculation would terminate too early with different layering strategies

## Checklist:


include/heuristic/HeuristicMapper.hpp
```diff
@@ -60,17 +63,24 @@ class HeuristicMapper : public Mapper {
...
+      costFixed = initCostFixed;
     }
 
     /**
      * @brief applies an in-place swap of 2 qubits in `qubits` and `locations`
      * of the node
      */
-    void applySWAP(const Edge& swap, Architecture& arch) {
+    void applySWAP(const Edge& swap, Architecture& arch,
+                   const std::vector<std::uint16_t> singleQubitGateMultiplicity,
```
burgholzer, line 79: Is this parameter not passed by (const) reference on purpose? This always incurs a copy.

include/heuristic/HeuristicMapper.hpp
```diff
@@ -60,17 +63,24 @@ class HeuristicMapper : public Mapper {
...
+                   bool                             considerFidelity) {
+      const auto& singleQubitFidelities = arch.getSingleQubitFidelities();
+      const auto& twoQubitFidelities    = arch.getFidelityTable();
```
burgholzer, line 82: I guess these statements could be moved into the if statement down below.

include/heuristic/HeuristicMapper.hpp
```diff
@@ -93,13 +103,45 @@ class HeuristicMapper : public Mapper {
...
+            (-(singleQubitGateMultiplicity.at(static_cast<std::size_t>(q2)) -
+               singleQubitGateMultiplicity.at(static_cast<std::size_t>(q1))) *
+                 log2(singleQubitFidelities.at(swap.first)) -
+             (singleQubitGateMultiplicity.at(static_cast<std::size_t>(q1)) -
+              singleQubitGateMultiplicity.at(static_cast<std::size_t>(q2))) *
+                 log2(singleQubitFidelities.at(swap.second)));
+        // adding cost oft the swap gate
+        if (arch.bidirectional()) {
+          costFixed +=
+              -3 * log2(twoQubitFidelities.at(swap.first).at(swap.second));
```
burgholzer, line 120: Prefer to use `std::log2` instead of `log2` (which is, technically, not guaranteed to be defined).

include/heuristic/HeuristicMapper.hpp
```diff
@@ -93,13 +103,45 @@ class HeuristicMapper : public Mapper {
...
         throw QMAPException("Something wrong in applySWAP.");
       }
+
+      if (considerFidelity) {
+        // accounting for fidelity difference of single qubit gates (two qubit
+        // gates are handled in the heuristic)
+        costFixed +=
+            (-(singleQubitGateMultiplicity.at(static_cast<std::size_t>(q2)) -
+               singleQubitGateMultiplicity.at(static_cast<std::size_t>(q1))) *
+                 log2(singleQubitFidelities.at(swap.first)) -
```
burgholzer, line 113: Would it be helpful to straight up store log-fidelities? Avoids the constant computation here.

include/heuristic/HeuristicMapper.hpp
```diff
@@ -93,13 +103,45 @@ class HeuristicMapper : public Mapper {
...
+             (singleQubitGateMultiplicity.at(static_cast<std::size_t>(q1)) -
+              singleQubitGateMultiplicity.at(static_cast<std::size_t>(q2))) *
+                 log2(singleQubitFidelities.at(swap.second)));
+        // adding cost oft the swap gate
```
burgholzer, line 117:
```suggestion
        // adding cost of the swap gate
```

include/heuristic/HeuristicMapper.hpp
```diff
@@ -93,13 +103,45 @@ class HeuristicMapper : public Mapper {
...
 
     /**
      * @brief applies an in-place teleportation of 2 qubits in `qubits` and
      * `locations` of the node
      */
-    void applyTeleportation(const Edge& swap, Architecture& arch) {
+    void applyTeleportation(const Edge& swap, Architecture& arch,
+                            bool considerFidelity) {
+      // const auto& singleQubitFidelities = arch.getSingleQubitFidelities();
+      // const auto& twoQubitFidelities = arch.getFidelityTable();
```
burgholzer: Could also just be moved down into the other code block to have it at the proper location.

include/heuristic/HeuristicMapper.hpp
```diff
@@ -151,6 +193,99 @@ class HeuristicMapper : public Mapper {
...
+     * @param arch the architecture for calculating distances between physical
+     * qubits and supplying qubit information such as fidelity
+     * @param singleQubitGateMultiplicity vector containing the number of gates
+     * acting on each logical qubit
+     * @param considerFidelity controls if qubit fidelity should be taken into
+     * account
+     */
+    void recalculateFixedCost(
+        const Architecture&              arch,
+        const std::vector<std::uint16_t> singleQubitGateMultiplicity,
```
burgholzer, line 235: Same question as above: why no const reference?

include/heuristic/HeuristicMapper.hpp
```diff
@@ -151,6 +193,99 @@ class HeuristicMapper : public Mapper {
...
+        // adding costs of single qubit gates
+        for (std::uint16_t i = 0U; i < arch.getNqubits(); ++i) {
+          if (singleQubitGateMultiplicity.at(i) == 0) {
+            continue;
+          }
+          costFixed += -singleQubitGateMultiplicity.at(i) *
+                       log2(singleQubitFidelities.at(
+                           static_cast<std::size_t>(locations.at(i))));
+        }
+        // adding cost oft the swap gates
```
burgholzer, line 251:
```suggestion
        // adding cost of the swap gates
```

include/heuristic/HeuristicMapper.hpp
```diff
@@ -151,6 +193,99 @@ class HeuristicMapper : public Mapper {
...
+            if (swap.op == qc::SWAP) {
+              if (arch.bidirectional()) {
+                costFixed +=
+                    -3 *
+                    log2(twoQubitFidelities.at(swap.first).at(swap.second));
+              } else {
+                costFixed += (-3 * log2(twoQubitFidelities.at(swap.first)
+                                            .at(swap.second)) -
+                              2 * log2(singleQubitFidelities.at(swap.first)) -
+                              2 * log2(singleQubitFidelities.at(swap.second)));
```
burgholzer, line 263: This looks like Code duplication with the function above. Consider extracting that into a function.

include/heuristic/HeuristicMapper.hpp
```diff
@@ -151,6 +193,99 @@ class HeuristicMapper : public Mapper {
...
+     * swaps
+     *
+     * @param arch the architecture for calculating distances between physical
+     * qubits and supplying qubit information such as fidelity
+     * @param singleQubitGateMultiplicity vector containing the number of gates
+     * acting on each logical qubit
+     * @param considerFidelity controls if qubit fidelity should be taken into
+     * account
+     */
+    void recalculateFixedCost(
```
burgholzer, line 233: I'd suggest moving the definition of the function to the source file. In order to keep the header file "clean". Probably applies to some of the other functions as well.

include/heuristic/HeuristicMapper.hpp
```diff
@@ -294,7 +542,9 @@ class HeuristicMapper : public Mapper {
    * @param layer index of current circuit layer
    */
   void expandNode(const std::vector<std::uint16_t>& consideredQubits,
-                  Node& node, std::size_t layer);
+                  Node& node, std::size_t layer,
+                  const std::vector<std::uint16_t> singleQubitGateMultiplicity,
+                  const EdgeMultiplicity           twoQubitGateMultiplicity);
```
burgholzer, line 544: Both of these could probably be const references. Please also check for all other functions taking these new parameters.

include/utils.hpp
```diff
@@ -61,12 +61,13 @@ class Dijkstra {
   };
 
   static void buildTable(std::uint16_t n, const CouplingMap& couplingMap,
-                         Matrix& distanceTable,
+                         Matrix<double>&      distanceTable,
+                         const Matrix<double> edgeWeights,
```
burgholzer, line 65: Shouldn't this also be a const reference?

include/utils.hpp
```diff
@@ -61,12 +61,13 @@ class Dijkstra {
...
                          const std::function<double(const Node&)>& cost);
 
 protected:
   static void dijkstra(const CouplingMap& couplingMap, std::vector<Node>& nodes,
-                       std::uint16_t start);
+                       std::uint16_t start, const Matrix<double> edgeWeights);
```
burgholzer, line 70: Same as above

src/Architecture.cpp
```diff
@@ -185,20 +189,43 @@ Architecture::Architecture(const std::uint16_t nQ, const CouplingMap& cm,
 }
 
 void Architecture::createDistanceTable() {
+  Matrix<double> edgeWeights(nqubits, std::vector<double>(nqubits, INFINITY));
```
burgholzer, line 192: Prefer using the appropriate `max` value from `std::numeric_limits` instead of INFINITY.

src/Architecture.cpp
```diff
@@ -185,20 +189,43 @@ Architecture::Architecture(const std::uint16_t nQ, const CouplingMap& cm,
...
   }
 
-  if (isBidirectional) {
-    Dijkstra::buildTable(nqubits, couplingMap, distanceTable,
-                         Architecture::costHeuristicBidirectional);
-  } else {
-    Dijkstra::buildTable(nqubits, couplingMap, distanceTable,
-                         Architecture::costHeuristicUnidirectional);
+  Dijkstra::buildTable(nqubits, couplingMap, distanceTable, edgeWeights,
+                       Architecture::dijkstraNodeToCostNonFidelity);
```
burgholzer, line 201: Why is there no distinction between directionality here anymore?

src/Architecture.cpp
```diff
@@ -185,20 +189,43 @@ Architecture::Architecture(const std::uint16_t nQ, const CouplingMap& cm,
...
+  Matrix<double> edgeWeights(nqubits, std::vector<double>(nqubits, INFINITY));
+  for (const auto& edge : couplingMap) {
+    edgeWeights.at(edge.first).at(edge.second) =
+        -3 * log2(fidelityTable.at(edge.first).at(edge.second));
+    if (couplingMap.find({edge.second, edge.first}) == couplingMap.end()) {
+      edgeWeights.at(edge.second).at(edge.first) =
+          -3 * log2(fidelityTable.at(edge.first).at(edge.second)) -
+          2 * log2(singleQubitFidelities.at(edge.first)) -
+          2 * log2(singleQubitFidelities.at(edge.second));
+    }
```
burgholzer, line 224: I have seen Code like this somewhere before up in the code. Maybe also something to extract and reuse.

src/heuristic/HeuristicMapper.cpp
```diff
@@ -16,6 +16,27 @@ void HeuristicMapper::map(const Configuration& configuration) {
               << " not suitable for heuristic mapper!" << std::endl;
     return;
   }
+  if (config.considerFidelity &&
+      architecture.getSingleQubitFidelities().size() == 0) {
```
burgholzer, line 20: Use the `empty` method instead of a size is zero check.

src/heuristic/HeuristicMapper.cpp
```diff
@@ -16,6 +16,27 @@ void HeuristicMapper::map(const Configuration& configuration) {
...
+    std::cerr << "No calibration data available for this architecture! "
+              << "Performing mapping without considering fidelity."
+              << std::endl;
+    config.considerFidelity = false;
+  }
+  if (config.considerFidelity && config.lookahead) {
+    std::cerr << "Lookahead is not yet supported for heuristic mapper using "
+                 "fidelity-aware mapping!"
```
burgholzer, line 28: This should probably state that mapping is still performed, but without lookahead similar to the above warning.

src/heuristic/HeuristicMapper.cpp
```diff
@@ -16,6 +16,27 @@ void HeuristicMapper::map(const Configuration& configuration) {
...
+              << std::endl;
+    config.lookahead = false;
+  }
+  if (config.considerFidelity &&
+      config.initialLayout == InitialLayout::Dynamic) {
+    std::cerr << "Initial layout strategy " << toString(config.initialLayout)
+              << " not yet supported for heuristic mapper using fidelity-aware "
+                 "mapping!"
```
burgholzer, line 36: This should probably state that the mapping is aborted.

src/heuristic/HeuristicMapper.cpp
```diff
@@ -306,6 +339,7 @@ void HeuristicMapper::mapUnmappedGates(
     std::size_t layer, std::vector<std::uint16_t>& consideredQubits) {
   for (const auto& gate : layers.at(layer)) {
     if (gate.singleQubit()) {
+      // TODO: if considerFidelity probably necessary to also map single qubits
```
burgholzer, line 342: Do you plan to address this Todo within this PR or would you rather prefer to do that as a follow up?

src/heuristic/HeuristicMapper.cpp
```diff
@@ -438,45 +507,48 @@ void HeuristicMapper::expandNode(
...
-      }
-      if (g.second == node.locations.at(qc.getNqubits() + i + 1) &&
-          g.first != node.locations.at(qc.getNqubits() + i)) {
-        e.first = g.first;
-        e.second =
-            static_cast<std::uint16_t>(node.locations.at(qc.getNqubits() + i));
-        architecture.getCurrentTeleportations().insert(e);
-        perms.insert(e);
+  // for now teleportation is not supported for noise-aware mapping
+  if (!results.config.considerFidelity) {
```
burgholzer, line 511: Instead of wrapping everything in an if and increasing the indentation level, prefer to just return early for the not-supported case.

test/test_heuristic.cpp
```diff
@@ -232,3 +232,66 @@ TEST_P(HeuristicTest20QTeleport, Teleportation) {
...
+
+TEST_P(HeuristicTest5QFidelity, NoFidelity) {
+  Configuration settings{};
+  settings.initialLayout    = InitialLayout::Static;
+  settings.considerFidelity = true;
+  nonFidelityMapper->map(settings);
+  nonFidelityMapper->dumpResult(GetParam() +
+                                "_heuristic_london_nofidelity.qasm");
+  nonFidelityMapper->printResult(std::cout);
+  SUCCEED() << "Mapping successful";
```
burgholzer, line 296: I like these kind of tests in principle, as they at least show that _something_ is working properly.
However, it would be great to also add some tests that really scrutinize the implementation and see whether the "correct" result has been obtained.
That also holds for these fidelity table methods and everything else introduced in this PR.

burgholzer (changes requested):
Many thanks for this! Already looks great.
I just found some time to go through mostly everything. Haven't scrutinized all aspects of the PR yet since I am just on my phone in Tokyo, but I left you some pointers to things that I believe could be improved.
Also fixing the remaining clang-tidy warnings (you can check them by clicking on the CI run on GitHub) would be great. Some of them might even be stuff I annotated in the comments.
Many thanks!

include/heuristic/HeuristicMapper.hpp
EliasLF: Right, thank you. This is going to be fixed in the next commit.

include/heuristic/HeuristicMapper.hpp
EliasLF: Right, thank you. Will also be fixed in the next commit.

include/heuristic/HeuristicMapper.hpp
EliasLF: Well it would introduce the need for 2 additional member data structures inside `Architecture` (since raw fidelity values are already in use in some methods of `Architecture` itself, so existing `singleQubitFidelities` and `fidelityTable` cannot just replaced with log versions). 
But considering the number of calls to log2 here, I think you are right. Will also be implemented in a following commit on this PR. 
Thank you

include/heuristic/HeuristicMapper.hpp
EliasLF: Thank you, will be fixed in the next commit.

include/heuristic/HeuristicMapper.hpp
```diff
@@ -93,13 +103,45 @@ class HeuristicMapper : public Mapper {
...
+                 log2(singleQubitFidelities.at(swap.second)));
+        // adding cost oft the swap gate
+        if (arch.bidirectional()) {
+          costFixed +=
+              -3 * log2(twoQubitFidelities.at(swap.first).at(swap.second));
+        } else {
+          costFixed +=
+              (-3 * log2(twoQubitFidelities.at(swap.first).at(swap.second)) -
+               2 * log2(singleQubitFidelities.at(swap.first)) -
+               2 * log2(singleQubitFidelities.at(swap.second)));
```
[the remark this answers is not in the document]
EliasLF: You are absolutely right here. I just realised, that this PR could use some additional comments in a few other places, too. Coming up in a future commit. Thank you!

include/heuristic/HeuristicMapper.hpp
```diff
@@ -151,6 +193,99 @@ class HeuristicMapper : public Mapper {
...
+      const auto& twoQubitFidelities    = arch.getFidelityTable();
+
+      costFixed = 0;
+      if (considerFidelity) {
+        // adding costs of single qubit gates
+        for (std::uint16_t i = 0U; i < arch.getNqubits(); ++i) {
+          if (singleQubitGateMultiplicity.at(i) == 0) {
+            continue;
+          }
+          costFixed += -singleQubitGateMultiplicity.at(i) *
```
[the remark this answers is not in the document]
EliasLF: This was a (poor) attempt of increasing readability. As I am using the -log2 as fidelity-aware cost everywhere, and I thought it would be more obvious to be just a negating sign of the log resulting overall in an addition of a positive cost.
However, just a comment would probably be more helpful.

src/heuristic/HeuristicMapper.cpp
EliasLF: Returning early would not result in the same behavior here, as I just wrapped the part of the method, which added teleportation qubits into `perms` (which admittedly is a large chunk of the function), but the execution flow would still continue below that with the usual mapping. However, I instead set `configuration.teleportationQubits = 0` now at the beginning of `HeuristicMapper::map` with an accompanying warning message, which is cleaner anyway.

test/test_heuristic.cpp
EliasLF: I agree. I'm going to implement some further more rigorous tests for a future commit

src/Architecture.cpp
EliasLF: This is actually interesting and also caught me a little by surprise at first. 

Currently (on the main git branch) the dijkstra implementation just counts the minimum number of edges from one qubit to another without regard of edge directionality (in extreme cases this could btw even lead to a wrong minimal path, as it could in fact be cheaper to take one more forward edge instead of a pure back-edge path, resulting in an overestimation of the cost). 
It then takes the number of `(steps-1)*COST_BIDIRECTIONAL_SWAP` resp. `(steps-1)*COST_UNIDIRECTIONAL_SWAP` (which again could break admissibility by overestimation in the unidirectional case, if I'm not mistaken). In case the last edge was a back-edge in the unidirectional case an additional `COST_DIRECTION_REVERSE` is added.
From this I gathered that the current implementation places the actual CNOT gate on the last edge and already accounts for a potential reversal of that gate.

In my implementation dijkstra is supplied with the actual cost of each swap via edge weights (these are set appropriately in the bidirectional and unidirectional case respectively). The resulting cost of dijkstra is therefore already the full swap cost. However, since dijkstra again finds only the full path and we want to place our CNOT on the last edge, we need to subtract the last swap:
In the bidirectional case (or on a forward edge): `cost - COST_BIDIRECTIONAL_SWAP`
On a back-edge: `cost - COST_UNIDIRECTIONAL_SWAP`. Since it is a back-edge, however, we need to add `COST_DIRECTION_REVERSE` again, resulting in `cost - COST_UNIDIRECTIONAL_SWAP + COST_DIRECTION_REVERSE` which is of course equivalent to `cost - COST_BIDIRECTIONAL_SWAP`
Therefore removing the last swap is equivalent in both cases (as we only want to remove the CNOTs of the swap), and the directionality of all edges in-between is accounted for (correctly) inside `Dijkstra::dijkstra`.

src/heuristic/HeuristicMapper.cpp
EliasLF: I think I would tackle the lookahead problem next, as I feel the interaction between layers is even more integral.
Only after that am I planning on integrating the noise-aware approach with dynamic initial layouts.
How I should split those into individual PRs is completely up to you, though. Before and after each of these steps, the mapping should at least be working. However, I think the full potential of the noise-aware approach can only be utilized once lookahead and dynamic initial layouts are both implemented. I don't know which of those two is a sufficient or preferable state for integration into the main branch.

src/Architecture.cpp
burgholzer: Nice catch and thanks for the analysis! Now that makes a lot more sense. It need not be that detailed, but maybe a small comment about this would not hurt.

src/heuristic/HeuristicMapper.cpp
burgholzer: I don't think I have a real preference for the particular steps towards achieving the grand goal of noise-aware mapping.
Although I agree that a working mapper should result from all individual contributions. Other than that, I typically prefer small PRs to huge ones, so splitting the whole contribution up would make sense.
Feel free to proceed as you see fit. 
In the end, I'll cut a release once everything is out.

test/test_heuristic.cpp
burgholzer: I'll leave this conversation open then until such a commit has been added in order to keep track of it.

src/heuristic/HeuristicMapper.cpp
burgholzer: Ah, missed that on the phone and thought the function ended at that point. Sorry for the confusion.
Your change makes sense 👍🏼

include/heuristic/HeuristicMapper.hpp
burgholzer: These additional data structures should probably remain really small even for larger architectures, so I think their overhead is justified.
If that turns out to not be the case, the occurrences where the fidelities are used in a non-log fashion could probably be refactored.

src/Architecture.cpp
EliasLF: I'm very sorry, I just realized a mistake in my reasoning there. In the unidirectional case with a forward-edge in the last step the above would actually not hold, as in this case it would still be necessary to apply a unidirectional swap, but if just executing one CNOT in the direction of the edge, we would not need the CNOT reversal. Therefore we would have `cost - COST_UNIDIRECTIONAL_SWAP`.
To solve this confusion once and for all, I now just introduced a new field in the Dijkstra nodes, which keeps track of the cost without the last swap. Of course this could also have been solved by looking at `Architecture::bidirectional()` similar to the (still inaccurate) solution before, but seeing as we just want the swap cost without the last swap it seems easier to just not add the last swap cost in the first place instead of inferring what it was and subtracting it after the fact.

burgholzer: Alright. Sounds like a good solution to me! I'll look at it in detail later on.

burgholzer:
@EliasLF https://github.com/cda-tum/qmap/pull/240 might be interesting for you.

PR #240, opened 2023-02-16
Fix disjoint layer

lsschmid:
## Description

In the process of creating layers using the disjoint qubit strategy, multiple gates acting on the same qubit ended up in the same layer. As a result the heuristic mapping process ended up in an infinite looping searching for a possible mapping.

Problem:
When the previous gate was appended to a existing layer, the **lastLayer** variable was not updated correctly.

Fixes # (issue)
- fixed creation of disjoint layers

## Checklist:


burgholzer (approved):
Many thanks for catching this. What a stupid mistake 🤦🏼‍♂️

PR #247, opened 2023-02-25
fix disjoint qubit layering

EliasLF:
## Description

Currently, the layering strategy `DisjointQubits` always inserts gates in the last layer. By instead inserting it one layer after the first layer (from the back of the circuit), in which it would intersect with the qubit set of a previous gate, we achieve a still valid layering with less layers. Visually speaking, we move each gate as far left as possible without intersecting other gates.

The following visualizations illustrate the status-quo and the new solution proposed in this PR on the following circuit:
circuit_ordered
The numbers next to the gates indicate the gate order in the circuit.

Currently, the following layering strategy is implemented (red dots indicate the state of `laystLayer` for each qubit after inserting the gate):
current

The proposed implementation:
lukas

## Checklist:

### DIFF
--- a/src/Mapper.cpp
+++ b/src/Mapper.cpp
@@ -42,6 +42,7 @@ void Mapper::processDisjointQubitLayer(
     } else {
       layer = *lastLayer.at(target) + 1;
     }
+    lastLayer.at(target) = layer;
   } else {
     if (!lastLayer.at(*control).has_value() &&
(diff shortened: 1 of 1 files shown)


burgholzer (approved):
Thanks for another contribution here! And also thanks again for the nice visualisations! These certainly help a lot!

PR #248, opened 2023-02-25
adding Disjoint2qBlocks layering

EliasLF:
## Description

This PR implements a proposed new layering strategy `Disjoint2qBlocks`, which is equivalent to collecting all gates in 2Q-Blocks and then applying `DisjointQubit` layering.

The following illustrates this strategy on the circuit:
circuit_ordered

mine

## Checklist:


include/Mapper.hpp
```diff
@@ -124,6 +127,25 @@ class Mapper {
...
+   * except for gates acting on exactly matching qubit sets, which are
+   * collected in the same layer. Additionally also single qubit gates can be
+   * added to a layer already containing gates acting on that qubit.
+   * This approach is equivalent to collecting all gates in 2Q-Blocks and then
+   * applying DisjointQubitLayering.
+   *
+   * @param lastLayer the array storing the last layer each qubit is used in
+   * @param control the (potential) control qubit of the gate
+   * @param target the target qubit of the gate
+   * @param gate the gate to be added to the layerh
```
burgholzer, line 142:
```suggestion
   * @param gate the gate to be added to the layer
```

src/Mapper.cpp
```diff
@@ -70,6 +70,47 @@ void Mapper::processDisjointQubitLayer(
...
+
+  if (layers.size() <= layer) {
+    layers.emplace_back();
+  }
+  if (control.has_value()) {
+    layers.at(layer).emplace_back(*control, target, gate);
+  } else {
+    layers.at(layer).emplace_back(-1, target, gate);
+  }
+}
```
burgholzer, line 112: Just a thought: this shares quite some code with the disjoint qubits layering and only differs in a few places.
Would it be cleaner to add a flag to the existing function (something like `collect2qBlocks`) and control the behavior based on that flag?
Or would you think that makes the function too hard to understand and rather stick with the existing solution?

test/test_heuristic.cpp
```diff
@@ -159,7 +159,7 @@ TEST_P(HeuristicTest16Q, Disjoint) {
   Configuration settings{};
   settings.layering = Layering::DisjointQubits;
   ibmQX5Mapper->map(settings);
-  ibmQX5Mapper->dumpResult(GetParam() + "_heuristic_qx5_dynamic.qasm");
+  ibmQX5Mapper->dumpResult(GetParam() + "_heuristic_qx5_disjoint.qasm");
```
burgholzer, line 162: Good catch!

Would it make sense to also add a corresponding test for the heuristic mapper using the new layout strategy here?
Maybe this also helps to get the Coverage run in the green.

burgholzer (changes requested):
Thanks for the high-quality PR. 
I really like the idea and only have some small comments/suggestions/questions, which you can find below. Otherwise, this should be good to go.

src/Mapper.cpp
EliasLF: Thank you for the suggestion! I now combined the two methods

test/test_heuristic.cpp
EliasLF: I originally planned on doing so. (That is also why I catched that wrong file name)
However, using Disjoint2qBlocks layering that test would result in an extremely long running mapping for the circuit `qft_16` (`Heuristic/HeuristicTest16Q.Disjoint/qft_16`), because of a pre-existing bug with the ordering of A* search nodes. Since that bug is otherwise completely separate from the layering implementation, I just left that test out.

I am currently working on a solution for that bug. So maybe it makes sense to hold out on merging this PR (including the test) until after.

burgholzer: Yeah. I'll wait with the merge here until that bug is fixed. The changes here all look good. So once the bug is fixed and the tests are added, this should be good to go.

EliasLF: The tests are now added. Should be ready to be merged

burgholzer (approved):
LGTM 👍🏻

PR #264, opened 2023-03-16
🐛 Fix performance regression in exact mapper on directed architectures

### LINKED ISSUE #256: 🐛 Missmatches in current implmentation to publication
### mqt.qmap version

mqt-qmap 2.1.2 via pip

### OS

Ubuntu

### Python version

_No response_

### C++ compiler

_No response_

### Additional environment information

_No response_

### Description

Please find below an (incomplete) list of benchmarks where the current implementation of the **exact** mapping algorithm shows differences when reproducing the results of [Wille et al.](https://doi.org/10.1145/3316781.3317859) (IBM QX4 arch):

- 3_17_13.qasm
- 4gt11_83.qasm
- 4mod5-v0_19.qasm
- alu-v3_34.qasm

### Expected behavior

I would expect the same results as mentioned in the paper.

### How to Reproduce

Run listed benchmarks above with exact mapper for arch graph resembling IBM QX4 (asymmetric) connectivity and compare with results of publication.

burgholzer:
## Description

Fixes #256.
The underlying cause of the performance regression was an oversight in the addition of the SWAP limitation feature for the ASPDAC 2021 paper. The implementation did not properly handle directed architectures in various cases:
 - due to the unidirectional edges, the longest shortest path search would sometimes return paths that were too long. That alone wouldn't be a problem, but makes the performance suboptimal. This is fixed by making sure that the search for the longest shortest path considers edges in the coupling graph as bidirectional.
 - More critically, the coupling limit computation only computes the cost of moving any two qubits next to each other. However, in directed architectures, this might not be enough to make a two-qubit gate executable. The simplest case, where this can be seen is a directed two-qubit architecture and a circuit with two controlled-H gates that have flipped controls and targets. Obviously, a SWAP is required in that case, but the old coupling limit computation would result in a limit of zero. This is fixed by incrementing the coupling limit by one in case of directed architectures.
 - SWAP limits were activated even if `settings.swapReduction == SwapReduction::None`, which also led to the limit being set to zero. This was fixed by adjusting the condition for enabling swap reduction.
 - Last but not least, the solver optimization for directed architectures was not ideal and has been changed to using weighted clauses instead of separate minimisation calls.

## Checklist:

### DIFF
--- a/include/Architecture.hpp
+++ b/include/Architecture.hpp
@@ -15,6 +15,7 @@
 #include <map>
 #include <regex>
 #include <unordered_map>
+#include <unordered_set>
 #include <vector>
 
 constexpr std::uint8_t GATES_OF_BIDIRECTIONAL_SWAP  = 3U;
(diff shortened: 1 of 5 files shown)


burgholzer:
The combination of https://github.com/cda-tum/qmap/pull/264/commits/76b094d0e60c0295ac2145623de87e64933c6d3f and https://github.com/cda-tum/qmap/pull/264/commits/6f92dac15c70110b3a0f350b1424545b5893240a seems to have at least fixed part of the problem. Now only the test without the SWAP reduction strategy is failing.

PR #269, opened 2023-03-19
Dijkstra arbitrary weights

EliasLF:
## Description

Replacing fixed edge costs in Dijkstra implementation with an edge weight matrix. Thereby enabling future expansions (fidelity-aware distances, semi-directional architectures, etc.)

## Checklist:

### DIFF
--- a/include/Architecture.hpp
+++ b/include/Architecture.hpp
@@ -363,12 +363,8 @@ protected:
 
   void createDistanceTable();
   void createFidelityTable();
-  
-  static double dijkstraNodeToCostNonFidelityBidirectional(const Dijkstra::Node& node) {
-      return node.prevCost;
-  }
 
-  static double dijkstraNodeToCostNonFidelityUnidirectional(const Dijkstra::Node& node) {
+  static double dijkstraNodeToCostNonFidelity(const Dijkstra::Node& node) {
     // Dijkstra determines the minimal path cost from one physical qubit to 
     // another.  In the non-fidelity case we are only interested in swapping 2
     // logical qubits next to each other. Therefore the last swap cost has to 
--- a/include/Architecture.hpp
+++ b/include/Architecture.hpp
@@ -365,14 +365,15 @@ protected:
   void createFidelityTable();
 
   static double dijkstraNodeToCostNonFidelity(const Dijkstra::Node& node) {
-    // Dijkstra determines the minimal path cost from one physical qubit to 
+    // Dijkstra determines the minimal path cost from one physical qubit to
     // another.  In the non-fidelity case we are only interested in swapping 2
-    // logical qubits next to each other. Therefore the last swap cost has to 
-    // be ignored. That cost is stored in the field `node.prevCost` of each node.
+    // logical qubits next to each other. Therefore the last swap cost has to
+    // be ignored. That cost is stored in the field `node.prevCost` of each
+    // node.
     if (node.containsCorrectEdge) {
       return node.prevCost;
     } else {
-      // in case the last edge is a back-edge, we will need to reverse the CNOT, 
+      // in case the last edge is a back-edge, we will need to reverse the CNOT,
       // executed on that edge
       return node.prevCost + COST_DIRECTION_REVERSE;
     }
--- a/include/Architecture.hpp
+++ b/include/Architecture.hpp
@@ -363,22 +363,23 @@ protected:
 
   void createDistanceTable();
   void createFidelityTable();
-
-  static double costHeuristicBidirectional(const Dijkstra::Node& node) {
-    auto length = node.cost - 1;
-    if (node.containsCorrectEdge) {
-      return length * COST_BIDIRECTIONAL_SWAP;
-    }
-    throw QMAPException("In a bidrectional architecture it should not happen "
-                        "that a node does not contain the right edge.");
+  
+  static double dijkstraNodeToCostNonFidelityBidirectional(const Dijkstra::Node& node) {
+      return node.prevCost;
   }
 
-  static double costHeuristicUnidirectional(const Dijkstra::Node& node) {
-    auto length = node.cost - 1;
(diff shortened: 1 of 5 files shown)


include/Architecture.hpp
```diff
@@ -364,21 +364,18 @@ class Architecture {
   void createDistanceTable();
   void createFidelityTable();
 
-  static double costHeuristicBidirectional(const Dijkstra::Node& node) {
-    auto length = node.cost - 1;
+  static double dijkstraNodeToCostNonFidelity(const Dijkstra::Node& node) {
```
burgholzer, line 367: Just a very minor nit: the naming of this function is kind of non-ideal as there is no "fidelity" version. Might as well just call this function `dijkstraNodeToCost`.

test/test_general.cpp
```diff
@@ -52,3 +53,65 @@ TEST(General, TestLineParsing) {
...
+  /*
+            (6)
+        .----------.
+       \/          \/
+  0 <-> 1 --> 2 <-> 3
+    (1)   (2)   (3)
+  */
+  CouplingMap cm = {{0, 1}, {1, 0}, {1, 2}, {2, 3}, {3, 2}, {1, 3}, {3, 1}};
+  Matrix      edgeWeights = {
+      {0, 1, 0, 0}, {1, 0, 2, 6}, {0, 15, 0, 3}, {0, 6, 3, 0}};
```
burgholzer, line 67: Is the `15` in the third entry here on purpose? Maybe I just interpreted the comment above in the wrong way, but to me it matches every entry but this `15`.

test/test_general.cpp
```diff
@@ -52,3 +53,65 @@ TEST(General, TestLineParsing) {
...
+  EXPECT_EQ(distanceTable[2].size(), 4);
+  EXPECT_EQ(distanceTable[2][0], 10);
+  EXPECT_EQ(distanceTable[2][1], 9);
+  EXPECT_EQ(distanceTable[2][2], 0);
+  EXPECT_EQ(distanceTable[2][3], 3);
+  EXPECT_EQ(distanceTable[3].size(), 4);
+  EXPECT_EQ(distanceTable[3][0], 7);
+  EXPECT_EQ(distanceTable[3][1], 6);
+  EXPECT_EQ(distanceTable[3][2], 3);
+  EXPECT_EQ(distanceTable[3][3], 0);
```
burgholzer, line 91: I am not quite sure whether googletest is capable of that, but couldn't you write this in a much more compact fashion by specifying a `targetTable` (just like the edge weights matrix) and do `EXPECT_EQ(distanceTable, targetTable);`?

test/test_general.cpp
burgholzer, line 67: Both of these could be declared `const`.

test/test_general.cpp
```diff
@@ -52,3 +53,65 @@ TEST(General, TestLineParsing) {
...
+TEST(General, Dijkstra) {
+  /*
+            (6)
+        .----------.
+       \/          \/
+  0 <-> 1 --> 2 <-> 3
+    (1)   (2)   (3)
+  */
+  CouplingMap cm = {{0, 1}, {1, 0}, {1, 2}, {2, 3}, {3, 2}, {1, 3}, {3, 1}};
+  Matrix      edgeWeights = {
```
burgholzer, line 66: I think you can get around the clang-format error from the linter by inserting a newline in front of this line and reformatting. Always strange how two independent systems (clang-tidy locally, as well as pre-commit) say everything is alright, but a this system says no..

burgholzer (approved):
Many thanks. This looks good to me. I just have a couple of minor suggestions and small questions. Feel free to discard them. I also added a comment on how you might be able to circumvent the cpp-linter warning.

include/Architecture.hpp
EliasLF: I am going to submit another PR adding some preparatory functionality for fidelity-aware mapping (including a fidelity-aware version of this method) probably later this day. Since I did not know, if at that point this PR will already be merged, I just wanted to prevent merging conflicts with this naming. But if you want I can also rename it `dijkstraNodeToCost` for now.

test/test_general.cpp
EliasLF: Good catch. Thank you for the hint. But this is in fact on purpose. Our Dijkstra implementation generally ignores directionality of edges. The edge weight of `15` (which is higher than the sum of all other weights combined) kind of simulates a unidirectional edge. (Just to make sure that Dijkstra takes the correct path in that edge case)

EliasLF: Right, thank you

include/Architecture.hpp
burgholzer: I thought that this might be the case. I think I still have a slight preference for just removing the "NonFidelity" from the function name as it doesn't really add to the description of what the function does, whereas the fidelity version clearly states that it does something similar to the function here, but also considering fidelity.
However, this is definitely not a strong opinion. So feel free to leave the naming as it is.

test/test_general.cpp
burgholzer: That makes sense. Thanks for the explanation!

include/Architecture.hpp
EliasLF: Oh, I see, that is true. Alright, I changed it now.

test/test_general.cpp
EliasLF: I did not think that googletest supports deep collection comparisons, but it actually works. Thank you for the suggestion. The code is much cleaner now that way.

test/test_general.cpp
EliasLF: That worked, thank you! I have been slowly running out of ideas how to appease the linter there

burgholzer (approved):
LGTM 👍🏻 I'll merge once CI is satisfied.

test/test_general.cpp
burgholzer: Yeah. Sometimes it's a pain.. over time you develop a repertoire of tricks to fix these kind of issues 😌

PR #274, opened 2023-03-19
♻️ Refactor heuristic mapper

EliasLF:
## Description

- Replacing iterations over all gates in the heuristic with iterations over a collection counting gate multiplicities (i.e. the number of gates acting on each logical qubit)
- Moving large methods in `HeuristicMapper::Node` from hpp to cpp
- Moving all cost computations inside `HeuristicMapper::Node` methods, and only interacting with nodes via method calls

## Checklist:

### DIFF
--- a/include/heuristic/HeuristicMapper.hpp
+++ b/include/heuristic/HeuristicMapper.hpp
@@ -75,8 +75,8 @@ public:
     Node(const std::array<std::int16_t, MAX_DEVICE_QUBITS>& q,
          const std::array<std::int16_t, MAX_DEVICE_QUBITS>& loc,
          const std::vector<std::vector<Exchange>>&          sw            = {},
-         const double                                       initCostFixed = 0
-        ) : costFixed(initCostFixed) {
+         const double                                       initCostFixed = 0)
+        : costFixed(initCostFixed) {
       std::copy(q.begin(), q.end(), qubits.begin());
       std::copy(loc.begin(), loc.end(), locations.begin());
       std::copy(sw.begin(), sw.end(), std::back_inserter(swaps));
@@ -92,7 +92,9 @@ public:
     /**
      * @brief returns costFixed + lookaheadPenalty
      */
-    [[nodiscard]] double getTotalFixedCost() const { return costFixed + lookaheadPenalty; }
+    [[nodiscard]] double getTotalFixedCost() const {
+      return costFixed + lookaheadPenalty;
+    }
 
     /**
      * @brief applies an in-place swap of 2 qubits in `qubits` and `locations`
--- a/include/heuristic/HeuristicMapper.hpp
+++ b/include/heuristic/HeuristicMapper.hpp
@@ -8,7 +8,7 @@
 
 #pragma once
 
-struct pairHash {
+struct PairHash {
   template <class T1, class T2>
   std::size_t operator()(const std::pair<T1, T2>& p) const {
     auto h1 = std::hash<T1>{}(p.first);
--- a/include/heuristic/HeuristicMapper.hpp
+++ b/include/heuristic/HeuristicMapper.hpp
@@ -9,14 +9,14 @@
 #pragma once
 
 /**
- * number of two qubit gates acting on pairs of logical qubits in some layer
- * where the key entry corresponds to logical qubits pairs ({q1, q2}) and
- * the value entry to the number of gates acting on the pair in each direction
- * (the first number with control=q1, target=q2 and the second the reverse)
+ * number of two-qubit gates acting on pairs of logical qubits in some layer
+ * where the keys correspond to logical qubit pairs ({q1, q2}) and
+ * the values to the number of gates acting on a pair in each direction
+ * (the first number with control=q1, target=q2 and the second the reverse).
  *
- * e.g. with multiplicity {{0,1},{2,3}} there are 2 gates with logical
+ * e.g., with multiplicity {{0,1},{2,3}} there are 2 gates with logical
  * qubit 0 as control and qubit 1 as target, and 3 gates with 1 as control
- * and 0 as target
+ * and 0 as target.
  */
 using TwoQubitMultiplicity =
     std::map<Edge, std::pair<std::uint16_t, std::uint16_t>>;
--- a/include/heuristic/HeuristicMapper.hpp
+++ b/include/heuristic/HeuristicMapper.hpp
@@ -38,7 +38,7 @@ public:
    * swaps, mappings and costs
    */
   struct Node {
-    /** current fixed cost (for non-fidelity-aware mapping all cost of all swaps already added) */
+    /** current fixed cost (for non-fidelity-aware mapping cost of all swaps already added) */
     double costFixed = 0;
     /** heuristic cost expected for future swaps needed in current circuit layer
      */
--- a/include/heuristic/HeuristicMapper.hpp
+++ b/include/heuristic/HeuristicMapper.hpp
@@ -75,24 +75,24 @@ public:
     Node(const std::array<std::int16_t, MAX_DEVICE_QUBITS>& q,
          const std::array<std::int16_t, MAX_DEVICE_QUBITS>& loc,
          const std::vector<std::vector<Exchange>>&          sw            = {},
-         const double                                       initCostFixed = 0) {
+         const double                                       initCostFixed = 0
+        ) : costFixed(initCostFixed) {
       std::copy(q.begin(), q.end(), qubits.begin());
       std::copy(loc.begin(), loc.end(), locations.begin());
       std::copy(sw.begin(), sw.end(), std::back_inserter(swaps));
-      costFixed = initCostFixed;
     }
 
     /**
      * @brief returns costFixed + costHeur + lookaheadPenalty
      */
-    double getTotalCost() const {
+    [[nodiscard]] double getTotalCost() const {
       return costFixed + costHeur + lookaheadPenalty;
     }
 
     /**
      * @brief returns costFixed + lookaheadPenalty
      */
-    double getTotalFixedCost() const { return costFixed + lookaheadPenalty; }
+    [[nodiscard]] double getTotalFixedCost() const { return costFixed + lookaheadPenalty; }
 
     /**
      * @brief applies an in-place swap of 2 qubits in `qubits` and `locations`
--- a/include/heuristic/HeuristicMapper.hpp
+++ b/include/heuristic/HeuristicMapper.hpp
@@ -8,10 +8,23 @@
 
 #pragma once
 
+struct pair_hash {
+    template <class T1, class T2>
+    std::size_t operator () (const std::pair<T1,T2> &p) const {
+        auto h1 = std::hash<T1>{}(p.first);
+        auto h2 = std::hash<T2>{}(p.second);
+
+        if constexpr (sizeof(size_t) >= 8) {
+          return h1 ^ (h2 + 0x517cc1b727220a95 + (h2 << 6) + (h2 >> 2));
+        }
+        return h1 ^ (h2 + 0x9e3779b9 + (h2 << 6) + (h2 >> 2));
+    }
+};
+
 /**
  * number of two-qubit gates acting on pairs of logical qubits in some layer
- * where the keys correspond to logical qubit pairs ({q1, q2}) and
- * the values to the number of gates acting on a pair in each direction
+ * where the keys correspond to logical qubit pairs ({q1, q2}, with q1<=q2)
+ * and the values to the number of gates acting on a pair in each direction
  * (the first number with control=q1, target=q2 and the second the reverse).
  *
  * e.g., with multiplicity {{0,1},{2,3}} there are 2 gates with logical
@@ -19,7 +32,7 @@
  * and 0 as target.
  */
 using TwoQubitMultiplicity =
-    std::map<Edge, std::pair<std::uint16_t, std::uint16_t>>;
+    std::unordered_map<Edge, std::pair<std::uint16_t, std::uint16_t>, pair_hash>;
 
 class HeuristicMapper : public Mapper {
 public:
@@ -230,8 +243,8 @@ protected:
    * @param twoQubitGateMultiplicity number of two qubit gates acting on pairs
    * of logical qubits in the current layer
    */
-  void expandNode(const std::set<std::uint16_t>& consideredQubits, Node& node,
-                  std::size_t                 layer,
+  void expandNode(const std::unordered_set<std::uint16_t>& consideredQubits, 
+                  Node& node, std::size_t layer,
                   const TwoQubitMultiplicity& twoQubitGateMultiplicity);
 
   /**
--- a/include/heuristic/HeuristicMapper.hpp
+++ b/include/heuristic/HeuristicMapper.hpp
@@ -9,16 +9,16 @@
 #pragma once
 
 struct pair_hash {
-    template <class T1, class T2>
-    std::size_t operator () (const std::pair<T1,T2> &p) const {
-        auto h1 = std::hash<T1>{}(p.first);
-        auto h2 = std::hash<T2>{}(p.second);
+  template <class T1, class T2>
+  std::size_t operator()(const std::pair<T1, T2>& p) const {
+    auto h1 = std::hash<T1>{}(p.first);
+    auto h2 = std::hash<T2>{}(p.second);
 
-        if constexpr (sizeof(size_t) >= 8) {
-          return h1 ^ (h2 + 0x517cc1b727220a95 + (h2 << 6) + (h2 >> 2));
-        }
-        return h1 ^ (h2 + 0x9e3779b9 + (h2 << 6) + (h2 >> 2));
+    if constexpr (sizeof(size_t) >= 8) {
+      return h1 ^ (h2 + 0x517cc1b727220a95 + (h2 << 6) + (h2 >> 2));
     }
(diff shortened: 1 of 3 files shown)


include/heuristic/HeuristicMapper.hpp
```diff
@@ -8,6 +8,19 @@
...
+/**
+ * number of two qubit gates acting on pairs of logical qubits in some layer
+ * where the key entry corresponds to logical qubits pairs ({q1, q2}) and
+ * the value entry to the number of gates acting on the pair in each direction
+ * (the first number with control=q1, target=q2 and the second the reverse)
+ *
+ * e.g. with multiplicity {{0,1},{2,3}} there are 2 gates with logical
+ * qubit 0 as control and qubit 1 as target, and 3 gates with 1 as control
+ * and 0 as target
+ */
```
burgholzer, line 20:
```suggestion
/**
 * number of two-qubit gates acting on pairs of logical qubits in some layer
 * where the keys correspond to logical qubit pairs ({q1, q2}) and
 * the values to the number of gates acting on a pair in each direction
 * (the first number with control=q1, target=q2 and the second the reverse).
 *
 * e.g., with multiplicity {{0,1},{2,3}} there are 2 gates with logical
 * qubit 0 as control and qubit 1 as target, and 3 gates with 1 as control
 * and 0 as target.
 */
```

include/heuristic/HeuristicMapper.hpp
```diff
@@ -8,6 +8,19 @@
...
+using TwoQubitMultiplicity =
+    std::map<Edge, std::pair<std::uint16_t, std::uint16_t>>;
```
burgholzer, line 22: Is `map` here really preferred over `unordered_map` (=hash map)?

include/heuristic/HeuristicMapper.hpp
```diff
@@ -60,98 +74,48 @@ class HeuristicMapper : public Mapper {
...
      * @brief applies an in-place teleportation of 2 qubits in `qubits` and
      * `locations` of the node
      */
-    void applyTeleportation(const Edge& swap, Architecture& arch) {
-      const auto q1 = qubits.at(swap.first);
-      const auto q2 = qubits.at(swap.second);
-
-      qubits.at(swap.first)  = q2;
-      qubits.at(swap.second) = q1;
+    void applyTeleportation(const Edge& swap, Architecture& arch);
```
burgholzer, line 109: Shouldn't it be possible for `arch` to be a `const` reference?

src/heuristic/HeuristicMapper.cpp
```diff
@@ -301,21 +317,15 @@ void HeuristicMapper::createInitialMapping() {
...
 void HeuristicMapper::mapUnmappedGates(
-    std::size_t layer, std::vector<std::uint16_t>& consideredQubits) {
-  for (const auto& gate : layers.at(layer)) {
-    if (gate.singleQubit()) {
-      continue;
-    }
+    const TwoQubitMultiplicity& twoQubitGateMultiplicity) {
+  for (const auto& edgeMultiplicity : twoQubitGateMultiplicity) {
+    auto q1 = edgeMultiplicity.first.first;
+    auto q2 = edgeMultiplicity.first.second;
```
burgholzer, line 323: You could benefit from using structured bindings here. Something like
```suggestion
  for (const auto& [edge, multiplicity] : twoQubitGateMultiplicity) {
    const auto& [q1, q2] = edge;
```
You might require a `[[maybe_unused]]` for the `multiplicity` binding, or try to get around the unused warning by using `_` as a name instead, which should also work, i.e., 
```c++
  for (const auto& [edge, _] : twoQubitGateMultiplicity) {
    const auto& [q1, q2] = edge;
```

src/heuristic/HeuristicMapper.cpp
```diff
@@ -395,23 +400,46 @@ void HeuristicMapper::mapToMinDistance(const std::uint16_t source,
...
+        twoQubitGateMultiplicity[{q1, q2}] = {!reverse, reverse};
+      } else {
+        if (reverse) {
+          twoQubitGateMultiplicity[{q1, q2}].second++;
+        } else {
+          twoQubitGateMultiplicity[{q1, q2}].first++;
+        }
+      }
+    }
+  }
```
burgholzer, line 427: Just a suggestion. Probably missing some static_casts.
Might look a little cleaner. Not the biggest fan of ternary if's and the implicit conversion in L418 (although it looks quite elegant). I'll leave it up to you whether you want to adapt.
What I'd do anyway, is to pre-construct the `{q1, q2}` pair in your code.
```suggestion
  for (const auto& gate : layers.at(layer)) {
    if (!gate.singleQubit()) {
      consideredQubits.emplace(gate.control);
      consideredQubits.emplace(gate.target);
      if (gate.control >= gate.target) {
        const auto edge = std::pair{gate.target, gate.control};
        if (twoQubitGateMultiplicity.find(edge) == twoQubitGateMultiplicity.end()) {
          twoQubitGateMultiplicity[edge] = {0, 1};
        } else {
          twoQubitGateMultiplicity[edge].second++;
        }
      } else {
        const auto edge = std::pair{gate.control, gate.target};
        if (twoQubitGateMultiplicity.find(edge) == twoQubitGateMultiplicity.end()) {
          twoQubitGateMultiplicity[edge] = {1, 0};
        } else {
          twoQubitGateMultiplicity[edge].first++;
        }
      }
  }
```

src/heuristic/HeuristicMapper.cpp
```diff
@@ -395,23 +400,46 @@ void HeuristicMapper::mapToMinDistance(const std::uint16_t source,
 }
 
 HeuristicMapper::Node HeuristicMapper::aStarMap(size_t layer) {
-  std::vector<std::uint16_t> consideredQubits{};
-  Node                       node{};
+  std::set<std::uint16_t> consideredQubits{};
```
burgholzer, line 403: Is it important, that this set is ordered? Or would a `std::unordered_set` do here and be potentially better performing?

src/heuristic/HeuristicMapper.cpp
```diff
@@ -597,3 +615,162 @@ void HeuristicMapper::lookahead(const std::size_t      layer,
                                          // for better fidelity lookahead
   }
 }
+
+void HeuristicMapper::Node::applySWAP(const Edge& swap, Architecture& arch) {
+  nswaps++;
```
burgholzer, line 620:
```suggestion
  ++nswaps;
```
small nit. probably doesn't even make a difference because the compiler rewrites it anyway.

src/heuristic/HeuristicMapper.cpp
```diff
@@ -597,3 +615,162 @@ void HeuristicMapper::lookahead(const std::size_t      layer,
...
+  if (arch.getCouplingMap().find(swap) != arch.getCouplingMap().end() ||
+      arch.getCouplingMap().find(Edge{swap.second, swap.first}) !=
+          arch.getCouplingMap().end()) {
+    swaps.back().emplace_back(swap.first, swap.second, qc::SWAP);
+  } else {
+    throw QMAPException("Something wrong in applySWAP.");
+  }
+
+  if (arch.bidirectional()) {
+    costFixed += COST_BIDIRECTIONAL_SWAP;
```
burgholzer, line 646: Just a side comment, because I thought about this now. This cost of a SWAP should probably be something that comes from the Architecture class. That class should be able to tell how a SWAP gate is realized on itself.
The Architecture class should probably also be able to answer the question "Can I apply this gate to these qubits?" and the follow-up question "How much does this cost me?" (Thinking of direction reversal costs).
Probably nothing for this PR, but good to keep in mind.

src/heuristic/HeuristicMapper.cpp
```diff
@@ -597,3 +615,162 @@ void HeuristicMapper::lookahead(const std::size_t      layer,
...
+        costFixed += COST_TELEPORTATION;
+      }
+    }
+  }
+}
+
+void HeuristicMapper::Node::updateHeuristicCost(
+    const Architecture&         arch,
+    const TwoQubitMultiplicity& twoQubitGateMultiplicity,
+    bool                        admissibleHeuristic) {
```
burgholzer, line 731: Small nit: can be `const` here.

src/heuristic/HeuristicMapper.cpp
```diff
@@ -597,3 +615,162 @@ void HeuristicMapper::lookahead(const std::size_t      layer,
...
+  costHeur = 0.;
+  done     = true;
+
+  // iterating over all virtual qubit pairs, that share a gate on the
+  // current layer
+  for (const auto& edgeMultiplicity : twoQubitGateMultiplicity) {
+    const auto& q1                   = edgeMultiplicity.first.first;
+    const auto& q2                   = edgeMultiplicity.first.second;
+    const auto& straightMultiplicity = edgeMultiplicity.second.first;
+    const auto& reverseMultiplicity  = edgeMultiplicity.second.second;
```
burgholzer, line 741: Structured bindings make this a lot more expressive.
```suggestion
  for (const auto& [edge, multiplicity] : twoQubitGateMultiplicity) {
    const auto& [q1, q2] = edge;
    
    const auto& [straightMultiplicity, reverseMultiplicity] = multiplicity;
```

src/heuristic/HeuristicMapper.cpp
```diff
@@ -597,3 +615,162 @@ void HeuristicMapper::lookahead(const std::size_t      layer,
...
+    if (admissibleHeuristic) {
+      if (straightMultiplicity > 0) {
+        costHeur = std::max(costHeur, swapCostStraight);
+      }
+      if (reverseMultiplicity > 0) {
+        costHeur = std::max(costHeur, swapCostReverse);
+      }
+    } else {
+      costHeur += swapCostStraight * straightMultiplicity +
+                  swapCostReverse * reverseMultiplicity;
```
burgholzer, line 773: In most cases, this will be an overestimate, right? For example, a multiplicity `{2, 0}` will only require a single `swapStraight`, right?
At the moment, this resembles the state of the `main` branch. Could this be improved somehow?

burgholzer (changes requested):
Many thanks for this refactor. The changes make sense to me and result in cleaner code.
I still have a couple of suggestions and/or comments, which you can find down below. 
Most of these should be fairly easy to address.
As for the red CI check: would it make sense to add 1-2 additional tests that cover some of the changes?

include/heuristic/HeuristicMapper.hpp
EliasLF: Oh, I was not aware, that C++ uses ordered collections by default. Thank you for the suggestion

include/heuristic/HeuristicMapper.hpp
EliasLF: Well, I just copied that function from `hpp` to `cpp`. And it seems, that it is not a `const` reference for a reason, since `arch.getTeleportationQubits()` (which is called in `applyTeleportation`) is not a `const` function and throws an error, when it turned into one.

burgholzer: Right... thought that is was missing for a reason. In that case, nothing to do here 🤷🏽‍♂️

src/heuristic/HeuristicMapper.cpp
EliasLF: Great suggestion. Makes the code much cleaner. Thank you!

src/heuristic/HeuristicMapper.cpp
EliasLF: Much cleaner, thank you!

src/heuristic/HeuristicMapper.cpp
EliasLF: You mean `admissibleHeuristic`? Right, I made it `const` now

burgholzer: Jap, exactly.

src/heuristic/HeuristicMapper.cpp
EliasLF: Thank you!

src/heuristic/HeuristicMapper.cpp
EliasLF: That is true. I copied the behavior of the `main` branch exactly. I thought, that a pure code refactor PR probably should not change the fundamental functionality. In its current form the code in this PR yields exactly the same A* nodes (with exactly the same costs) just with some little performance boost.

But if you find it appropriate, we could also optimize the above already in this PR. I think the best solution would be `costHeur += std::max(swapCostStraight, swapCostReverse)`.

However, I also thought a little bit about the idea of improving the admissible heuristic using the fact that any moving qubit can save at most a number of swaps equal to the number of other (potentially) moving qubits, and am planning on making another PR on that once we discussed it. So we could also wait with the above change for that PR.

burgholzer: I agree that it's probably best to keep this as a pure refactoring PR.
The follow-up PR can then improve both heuristics.

I like you suggestion for the non-admissible part. Should be as straight forward as that.

As for the admissible part, the idea makes sense, but I guess it's easier to reason over some code in that regard.

src/heuristic/HeuristicMapper.cpp
EliasLF: I agree. Would be much cleaner to have some sort of getter function, with probably only negligible performance overhead.

EliasLF:
I now implemented a test for the cost calculations of `HeuristicMapper::Node`. This provides some further coverage for a few lines.

However, looking at the CodeCov Report, the remaining uncovered lines seem to be almost exclusively error messages for catching bugs in the implementation. As long as everything runs as intended, there is therefore no way of covering those error messages with tests.
(The rest is some additional console output for `verbosity=true`, and some checks for the existence of the reverse of an edge)

Furthermore, I have no idea why the MacOS tests time out (while other systems seem to run it fine)

test/test_heuristic.cpp
```diff
@@ -29,6 +29,51 @@ class HeuristicTest5Q : public testing::TestWithParam<std::string> {
...
+  TwoQubitMultiplicity multiplicity = {{{0, 1}, {5, 2}}, {{2, 3}, {0, 1}}};
+  std::array<std::int16_t, MAX_DEVICE_QUBITS> qubits    = {4, 3, 1, 2, 0};
+  std::array<std::int16_t, MAX_DEVICE_QUBITS> locations = {4, 2, 3, 1, 0};
+
+  std::vector<std::vector<Exchange>> swaps = {
+      {Exchange(0, 1, qc::OpType::Teleportation)},
+      {Exchange(1, 2, qc::OpType::SWAP)}};
+
+  HeuristicMapper::Node node(qubits, locations, swaps, 5.);
+  EXPECT_NEAR(node.costFixed, 5., 1e-6);
```
burgholzer, line 44: Tiny nit: you could extract that `1e-6` into a `tolerance` constant value.

burgholzer (changes requested):
Thanks again for the changes. I just have very, very small comments.
I am totally fine with the patch coverage being red here since, as you said, a lot of error detection has added and I am happy that it is not triggered ;o)

I don't quite understand why that macOS test is failing. But it is failing consistently. I restarted the job and it, again, failed all three times the CI tried to run it due to hitting the 5min timeout.
It kind of has to be one of the changes in this PR... The only thing that would really make sense to me, are the `map` -> `unordered_map` and `set` -> `unordered_set` changes. You could try reverting them and see if it helps. Maybe one of the methods actually depends on the number of qubits being sorted. Or the different sequence of qubits has changed the A* result. Just hypothesising..

include/heuristic/HeuristicMapper.hpp
```diff
@@ -8,18 +8,31 @@
...
+  std::size_t operator()(const std::pair<T1, T2>& p) const {
+    auto h1 = std::hash<T1>{}(p.first);
+    auto h2 = std::hash<T2>{}(p.second);
+
+    if constexpr (sizeof(size_t) >= 8) {
+      return h1 ^ (h2 + 0x517cc1b727220a95 + (h2 << 6) + (h2 >> 2));
+    }
+    return h1 ^ (h2 + 0x9e3779b9 + (h2 << 6) + (h2 >> 2));
+  }
+};
```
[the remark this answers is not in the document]
EliasLF: I thought so too but it actually does not

test/test_heuristic.cpp
EliasLF: Good idea, thanks

burgholzer:
> The `unordered_map` seems to have been the problem. However, I checked both versions on my Ubuntu system and there I do not see any significant change in mapping time (or in anything else). If anything, there is even a slight _increase_ in mapping time using the ordered map. 
> 
> So it really seems to be a system-specific issue. Unfortunately, I do not have a MacOS system at hand to further look into it. I just hope, it is not a sign of some fundamental bug, that (for now) only appears in those specific circumstances but might come up later elsewhere.

At least it's fixed now.
The slight increase in mapping time from the ordered map is kind of expected (which is why I suggested the unordered map in the first place).
I have a macOS system at home, so I could digging into it. However, I am quite busy with my dissertation at the moment. So I am not sure when I could find the time for that.
What about the following: let's merge this PR as it is (everything is ready, right?). In addition, let's create an issue that recalls the problems we were facing here as a reminder to look into it. Does that make sense?

burgholzer (approved):
LGTM. Many thanks for this nice refactor and also for creating the relevant issue.
Merging this now and coming back to the issue when I have some more time.

PR #278, opened 2023-03-25
Heuristic Benchmarking

EliasLF:
## Description

Implements some benchmarks into the heuristic mapper (both for the whole mapping process, and for each layer):
- number of generated nodes
- number of expanded nodes
- depth of the retrieved solution in the search tree (only for each layer, not globally for the whole mapping process)
- average branching factor (i.e. how many children does a node generate on average)
- effective branching factor (i.e. how many of those children on average were expanded before the solution was found; or in other words the branching factor, which a balanced, uniform tree would need to have, to contain the number of processed nodes (expanded + 1) with a depth equal to the solution depth)
- the average time spent on each expanded node

All of this will help keeping track of performance changes, when implementing new heuristics or updating the existing heuristic.

## Checklist:

### DIFF
--- a/include/MappingResults.hpp
+++ b/include/MappingResults.hpp
@@ -109,17 +109,6 @@ struct MappingResults {
       benchmark["time_per_node"] = heuristicBenchmark.timePerNode;
       benchmark["average_branching_factor"] = heuristicBenchmark.averageBranchingFactor;
       benchmark["effective_branching_factor"] = heuristicBenchmark.effectiveBranchingFactor;
-      /*benchmark["layers"] = nlohmann::json::array();
-      auto& layerBenchmark = benchmark["layers"];
-      for (const auto& layer : layerHeuristicBenchmark) {
-        auto& layerStats = layerBenchmark.emplace_back();
-        layerStats["expanded_nodes"] = layer.expandedNodes;
-        layerStats["generated_nodes"] = layer.generatedNodes;
-        layerStats["solution_depth"] = layer.solutionDepth;
-        layerStats["time_per_node"] = layer.timePerNode;
-        layerStats["average_branching_factor"] = layer.averageBranchingFactor;
-        layerStats["effective_branching_factor"] = layer.effectiveBranchingFactor;
-      }*/
     }
     stats["additional_gates"] =
         static_cast<std::make_signed_t<decltype(output.gates)>>(output.gates) -
--- a/include/MappingResults.hpp
+++ b/include/MappingResults.hpp
@@ -27,6 +27,15 @@ struct MappingResults {
     std::size_t directionReverse = 0;
     std::size_t teleportations   = 0;
   };
+  
+  struct HeuristicBenchmarkInfo {
+    std::size_t expandedNodes            = 0;
+    std::size_t generatedNodes           = 0;
+    std::size_t solutionDepth            = 0;
+    double      timePerNode              = 0.;
+    double      averageBranchingFactor   = 0.;
+    double      effectiveBranchingFactor = 0.;
+  };
 
   CircuitInfo input{};
 
@@ -40,16 +49,21 @@ struct MappingResults {
   std::string mappedCircuit{};
 
   std::string wcnf{};
+  
+  HeuristicBenchmarkInfo heuristicBenchmark{};
+  std::vector<HeuristicBenchmarkInfo> layerHeuristicBenchmark{};
 
   MappingResults()          = default;
   virtual ~MappingResults() = default;
 
   virtual void copyInput(const MappingResults& mappingResults) {
-    input        = mappingResults.input;
-    architecture = mappingResults.architecture;
-    config       = mappingResults.config;
-    output       = mappingResults.output;
-    wcnf         = mappingResults.wcnf;
+    input                   = mappingResults.input;
+    architecture            = mappingResults.architecture;
+    config                  = mappingResults.config;
+    output                  = mappingResults.output;
+    wcnf                    = mappingResults.wcnf;
+    heuristicBenchmark      = mappingResults.heuristicBenchmark;
+    layerHeuristicBenchmark = mappingResults.layerHeuristicBenchmark;
   }
 
   [[nodiscard]] std::string toString() const { return json().dump(2); }
@@ -89,6 +103,23 @@ struct MappingResults {
       }
     } else if (config.method == Method::Heuristic) {
       stats["teleportations"] = output.teleportations;
+      auto& benchmark = stats["benchmark"];
+      benchmark["expanded_nodes"] = heuristicBenchmark.expandedNodes;
+      benchmark["generated_nodes"] = heuristicBenchmark.generatedNodes;
+      benchmark["time_per_node"] = heuristicBenchmark.timePerNode;
+      benchmark["average_branching_factor"] = heuristicBenchmark.averageBranchingFactor;
+      benchmark["effective_branching_factor"] = heuristicBenchmark.effectiveBranchingFactor;
+      /*benchmark["layers"] = nlohmann::json::array();
+      auto& layerBenchmark = benchmark["layers"];
+      for (const auto& layer : layerHeuristicBenchmark) {
+        auto& layerStats = layerBenchmark.emplace_back();
+        layerStats["expanded_nodes"] = layer.expandedNodes;
+        layerStats["generated_nodes"] = layer.generatedNodes;
+        layerStats["solution_depth"] = layer.solutionDepth;
+        layerStats["time_per_node"] = layer.timePerNode;
+        layerStats["average_branching_factor"] = layer.averageBranchingFactor;
+        layerStats["effective_branching_factor"] = layer.effectiveBranchingFactor;
+      }*/
     }
     stats["additional_gates"] =
         static_cast<std::make_signed_t<decltype(output.gates)>>(output.gates) -
--- a/include/MappingResults.hpp
+++ b/include/MappingResults.hpp
@@ -27,7 +27,7 @@ struct MappingResults {
     std::size_t directionReverse = 0;
     std::size_t teleportations   = 0;
   };
-  
+
   struct HeuristicBenchmarkInfo {
     std::size_t expandedNodes            = 0;
     std::size_t generatedNodes           = 0;
@@ -49,8 +49,8 @@ struct MappingResults {
   std::string mappedCircuit{};
 
   std::string wcnf{};
-  
-  HeuristicBenchmarkInfo heuristicBenchmark{};
+
+  HeuristicBenchmarkInfo              heuristicBenchmark{};
   std::vector<HeuristicBenchmarkInfo> layerHeuristicBenchmark{};
 
   MappingResults()          = default;
@@ -102,13 +102,15 @@ struct MappingResults {
         stats["WCNF"] = wcnf;
       }
     } else if (config.method == Method::Heuristic) {
-      stats["teleportations"] = output.teleportations;
-      auto& benchmark = stats["benchmark"];
-      benchmark["expanded_nodes"] = heuristicBenchmark.expandedNodes;
+      stats["teleportations"]      = output.teleportations;
+      auto& benchmark              = stats["benchmark"];
+      benchmark["expanded_nodes"]  = heuristicBenchmark.expandedNodes;
       benchmark["generated_nodes"] = heuristicBenchmark.generatedNodes;
-      benchmark["time_per_node"] = heuristicBenchmark.timePerNode;
-      benchmark["average_branching_factor"] = heuristicBenchmark.averageBranchingFactor;
-      benchmark["effective_branching_factor"] = heuristicBenchmark.effectiveBranchingFactor;
+      benchmark["time_per_node"]   = heuristicBenchmark.timePerNode;
+      benchmark["average_branching_factor"] =
+          heuristicBenchmark.averageBranchingFactor;
+      benchmark["effective_branching_factor"] =
+          heuristicBenchmark.effectiveBranchingFactor;
     }
     stats["additional_gates"] =
         static_cast<std::make_signed_t<decltype(output.gates)>>(output.gates) -
--- a/include/heuristic/HeuristicMapper.hpp
+++ b/include/heuristic/HeuristicMapper.hpp
@@ -5,6 +5,7 @@
 
 #include "Mapper.hpp"
 #include "heuristic/UniquePriorityQueue.hpp"
+
 #include <cmath>
 
 #pragma once
@@ -79,9 +80,8 @@ public:
     Node() = default;
     Node(const std::array<std::int16_t, MAX_DEVICE_QUBITS>& q,
          const std::array<std::int16_t, MAX_DEVICE_QUBITS>& loc,
-         const std::vector<std::vector<Exchange>>&          sw            = {},
-         const double                                       initCostFixed = 0,
-         const std::size_t                                  searchDepth   = 0)
+         const std::vector<std::vector<Exchange>>&          sw = {},
+         const double initCostFixed = 0, const std::size_t searchDepth = 0)
         : costFixed(initCostFixed), depth(searchDepth) {
       std::copy(q.begin(), q.end(), qubits.begin());
       std::copy(loc.begin(), loc.end(), locations.begin());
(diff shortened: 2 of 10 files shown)


src/heuristic/HeuristicMapper.cpp
```diff
@@ -439,10 +457,34 @@ HeuristicMapper::Node HeuristicMapper::aStarMap(size_t layer) {
...
+  if (results.layerHeuristicBenchmark.back().expandedNodes > 0) {
+    results.layerHeuristicBenchmark.back().timePerNode =
+        diff.count() /
+        static_cast<double>(
+            results.layerHeuristicBenchmark.back().expandedNodes);
+    results.layerHeuristicBenchmark.back().averageBranchingFactor =
+        static_cast<double>(
+            results.layerHeuristicBenchmark.back().generatedNodes - 1) /
+        static_cast<double>(
+            results.layerHeuristicBenchmark.back().expandedNodes);
```
burgholzer, line 487: Again, a very small performance improvement (probably what the compiler does internally): 
Place a 
```c++
const auto& layerResults = results.layerHeuristicBenchmark.back()
```
at the beginning of the block use that variable to access the layer results to avoid the additional level of indirection caused by the repeated `back()` calls.

src/heuristic/HeuristicMapper.cpp
```diff
@@ -453,6 +495,12 @@ HeuristicMapper::Node HeuristicMapper::aStarMap(size_t layer) {
     nodes.pop();
   }
 
+  results.layerHeuristicBenchmark.back().solutionDepth = result.depth;
+  results.layerHeuristicBenchmark.back().effectiveBranchingFactor =
+      computeEffectiveBranchingRate(
+          results.layerHeuristicBenchmark.back().expandedNodes + 1,
+          result.depth);
```
burgholzer, line 502: As above. Probably wouldn't hurt to add a 
```c++
const auto& layerResults = results.layerHeuristicBenchmark.back()
```

test/test_heuristic.cpp
```diff
@@ -81,6 +81,84 @@ TEST(Functionality, NodeCostCalculation) {
...
+
+  qc::QuantumComputation qc{5, 5};
+  qc.x(2, qc::Control{4});
+  qc.x(1, qc::Control{3});
+  qc.x(1, qc::Control{4});
+
+  qc.barrier({0, 1, 2, 3, 4});
+  for (size_t i = 0; i < 5; ++i) {
+    qc.measure(static_cast<qc::Qubit>(i), i);
+  }
```
burgholzer, line 106: Just a side note: at some point, we should really add a `measure_all` convenience function to the `QuantumComputation` class in the QFR similar to what Qiskit offers.

test/test_heuristic.cpp
```diff
@@ -81,6 +81,84 @@ TEST(Functionality, NodeCostCalculation) {
...
+
+  const auto    mapper = std::make_unique<HeuristicMapper>(qc, architecture);
```
burgholzer, line 108: Also just a side note: I am fairly sure that you could just allocate a plain `HeuristicMapper` object without the additional level of indirection.

test/test_heuristic.cpp
```diff
@@ -81,6 +81,84 @@ TEST(Functionality, NodeCostCalculation) {
...
+  */
+
+  EXPECT_EQ(result.layerHeuristicBenchmark.size(), 2);
+  EXPECT_EQ(result.layerHeuristicBenchmark[0].solutionDepth, 1);
+  EXPECT_EQ(result.layerHeuristicBenchmark[0].generatedNodes, 6);
+  EXPECT_EQ(result.layerHeuristicBenchmark[0].expandedNodes, 1);
+  EXPECT_NEAR(result.layerHeuristicBenchmark[0].averageBranchingFactor, 5.,
+              tolerance);
+  EXPECT_NEAR(result.layerHeuristicBenchmark[0].effectiveBranchingFactor, 1.,
+              tolerance);
```
burgholzer, line 147: It doesn't really matter that much since it's only tests, but also here you could extract a
```c++
const auto& layerResults0 = result.layerHeuristicBenchmark[0];
```
and similar for the second layer.

test/test_heuristic.cpp
```diff
@@ -142,10 +220,16 @@ TEST_P(HeuristicTest5Q, Identity) {
   ibmqYorktownMapper->map(settings);
   ibmqYorktownMapper->dumpResult(GetParam() + "_heuristic_qx4_identity.qasm");
   ibmqYorktownMapper->printResult(std::cout);
+  auto& result = ibmqYorktownMapper->getResults();
+  EXPECT_LE(result.heuristicBenchmark.effectiveBranchingFactor,
+            result.heuristicBenchmark.averageBranchingFactor);
```
burgholzer, line 225: just for my understanding: this is just a sanity check, right?

test/test_heuristic.cpp
```diff
@@ -156,8 +240,14 @@ TEST_P(HeuristicTest5Q, Static) {
...
   ibmqLondonMapper->map(settings);
+  auto& result = ibmqYorktownMapper->getResults();
+  EXPECT_LE(result.heuristicBenchmark.effectiveBranchingFactor,
+            result.heuristicBenchmark.averageBranchingFactor);
   ibmqLondonMapper->dumpResult(GetParam() + "_heuristic_london_static.qasm");
   ibmqLondonMapper->printResult(std::cout);
+  result = ibmqLondonMapper->getResults();
+  EXPECT_LE(result.heuristicBenchmark.effectiveBranchingFactor,
+            result.heuristicBenchmark.averageBranchingFactor);
   SUCCEED() << "Mapping successful";
```
burgholzer, line 251: Looking at all the lines that needed to be changed here, we should really be thinking about refactoring these tests to become less redundant.
It must be possible to write slightly more complicated test fixtures that allow to unify these tests in some fashion.
(Probably nothing for this PR)

burgholzer (changes requested):
Many thanks. I love the additional information this provides.
You can find a few small, nit-picky suggestions in the comments down below.

One bigger thing: have you tested how much the information tracking/computation affects the performance? Considering runtime, but also memory. Would it make sense to "hide" these computation behind a "debug" or "monitoring" or "benchmarking" flag in the configuration and only track the metrics if the flag is enabled? Or is the overhead negligible anyway and we might as well include it all the time?
I guess one of my main concern comes from the `Results` object being passed to Python which requires translating the C++ vector of results to Python. If the vector is large, this could be costly.

include/heuristic/HeuristicMapper.hpp
```diff
@@ -263,6 +269,34 @@ class HeuristicMapper : public Mapper {
...
+      for (std::size_t i = 1; i <= solutionDepth; ++i) {
+        sum += std::pow(mid, i);
+      }
+      if (sum < static_cast<double>(nodesProcessed)) {
+        lower = mid;
+      } else {
+        upper = mid;
+      }
+    }
+    return (lower + upper) / 2.0;
```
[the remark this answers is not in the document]
EliasLF: Well, yes, the return value would still be valid in that case, but it would need to perform 1 more iteration. Since the true solution is always guaranteed to be between `lower` and `upper`, the best guess is always the mid-point between those two. Returning one of the boundary values would be just throwing away half the precision, which could be compensated by another iteration, but I see no reason to do that.

include/MappingResults.hpp
```diff
@@ -88,7 +102,15 @@ struct MappingResults {
...
-      stats["teleportations"] = output.teleportations;
+      stats["teleportations"]      = output.teleportations;
+      auto& benchmark              = stats["benchmark"];
+      benchmark["expanded_nodes"]  = heuristicBenchmark.expandedNodes;
+      benchmark["generated_nodes"] = heuristicBenchmark.generatedNodes;
+      benchmark["time_per_node"]   = heuristicBenchmark.timePerNode;
+      benchmark["average_branching_factor"] =
+          heuristicBenchmark.averageBranchingFactor;
+      benchmark["effective_branching_factor"] =
+          heuristicBenchmark.effectiveBranchingFactor;
```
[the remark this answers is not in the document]
EliasLF: As I also mentioned in the PR description above, `solutionDepth` is only being tracked for each layer individually. I believe a global solution depth does not make too much sense. 
Implementing it as the sum of all layer-wise solution depths would be misleading, as running multiple separate search instances is not the same as running one with the sum of their solution depths (would result in much larger trees and therefore different effective branching rates). 
Implementing it as an average of all layers would make more sense, but would require another data type (floating point instead of integer) and would also be slightly confusing as the only global value with an average, that is not weighted by search tree size (which itself also would not make too much sense for the solution depth).
So I just decided to omit the solution depth for the global case.

include/heuristic/HeuristicMapper.hpp
```diff
@@ -263,6 +269,34 @@ class HeuristicMapper : public Mapper {
...
+    // no closed-form solution for b*, so we use approximation via binary search
+    if (solutionDepth == 0) {
+      return 0.;
+    }
+    --nodesProcessed; // N - 1 = (b*)^d + (b*)^(d-1) + ... + (b*)^2 + b*
+    double lower = std::pow(static_cast<double>(nodesProcessed),
+                            1.0 / static_cast<double>(solutionDepth)) /
+                   static_cast<double>(solutionDepth);
+    double upper = std::pow(static_cast<double>(nodesProcessed),
+                            1.0 / static_cast<double>(solutionDepth));
```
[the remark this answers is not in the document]
EliasLF: Thank you!

src/heuristic/HeuristicMapper.cpp
```diff
@@ -439,10 +457,34 @@ HeuristicMapper::Node HeuristicMapper::aStarMap(size_t layer) {
...
 
+  const auto start = std::chrono::steady_clock::now();
+
   while (!nodes.top().done) {
     Node current = nodes.top();
     nodes.pop();
     expandNode(consideredQubits, current, layer, twoQubitGateMultiplicity);
+
+    ++results.heuristicBenchmark.expandedNodes;
+    ++results.layerHeuristicBenchmark.back().expandedNodes;
```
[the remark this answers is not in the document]
EliasLF: Right, thank you!

test/test_heuristic.cpp
EliasLF: Yes, definitely. Especially since the mapper will initially not map any qubits that are not used nor measured (which makes sense for practical settings, but is often inconvenient during testing) So measuring all qubits is a nice work-around to force the mapper to map all qubits in the beginning according to the policy `Configuration::initialLayout`.

include/MappingResults.hpp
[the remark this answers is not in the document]
burgholzer: Thanks for the explanation! That makes total sense. Missed that when I looked over the PR.

include/heuristic/HeuristicMapper.hpp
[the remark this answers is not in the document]
burgholzer: Another good point. In that case, I am totally fine with the solution here.

test/test_heuristic.cpp
EliasLF: Well, I learned to always use `std::make_unique` if there is no good reason for not doing so (due to exception safety and such). But I have to admit that I am not an expert in such details of C++. So if you recommend using a direct instantiation here, I am going to change it.

test/test_heuristic.cpp
EliasLF: Yes exactly, if the effective branching rate is higher than the average branching rate, something went wrong in the benchmarking. On the one hand, adding those lines was just a large number of "free tests" for the benchmarking (the two branching factors also rely on all other benchmark values except for `timePerNode`), and on the other hand, if something went wrong there one should know, that they should not trust the benchmarking results of those tests.

test/test_heuristic.cpp
EliasLF: You are probably right, but I am really no expert on Google Test. So maybe we can find someone else for that, who is more familiar with the framework.

src/heuristic/HeuristicMapper.cpp
```diff
@@ -138,6 +139,21 @@ void HeuristicMapper::map(const Configuration& configuration) {
...
+        static_cast<double>(results.heuristicBenchmark.expandedNodes);
+    results.heuristicBenchmark.averageBranchingFactor =
+        static_cast<double>(results.heuristicBenchmark.generatedNodes -
+                            layers.size()) /
+        static_cast<double>(results.heuristicBenchmark.expandedNodes);
+    for (const auto& layer : results.layerHeuristicBenchmark) {
+      results.heuristicBenchmark.effectiveBranchingFactor +=
+          layer.effectiveBranchingFactor *
+          (static_cast<double>(layer.expandedNodes) /
+           static_cast<double>(results.heuristicBenchmark.expandedNodes));
```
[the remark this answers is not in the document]
EliasLF: I also might be mistaken on this, but I think the resulting precision might be higher this way. Since `expandedNodes` are usually rather large numbers (both in the global and the layer-wise case), I thought it might be beneficial to do that division before the multiplication. However, truthfully, I do not even know if the compiler adheres to the order of operations, that I tried to suggest via the parentheses.

EliasLF:
> One bigger thing: have you tested how much the information tracking/computation affects the performance? Considering runtime, but also memory. Would it make sense to "hide" these computation behind a "debug" or "monitoring" or "benchmarking" flag in the configuration and only track the metrics if the flag is enabled? Or is the overhead negligible anyway and we might as well include it all the time? I guess one of my main concern comes from the `Results` object being passed to Python which requires translating the C++ vector of results to Python. If the vector is large, this could be costly.

Oh, I did not even think about Python here. Using C++ directly, I did not see any significant change in mapping time (however, I also did not check memory usage, though I do not think there is any potential for a huge memory overhead). But I guess it cannot hurt to hide the feature behind a debugging flag in the configuration. Thank you for the suggestion!

test/test_heuristic.cpp
burgholzer: Right. I'll put it on my list of todos 👍🏻

src/heuristic/HeuristicMapper.cpp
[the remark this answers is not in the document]
burgholzer: Although I am fairly sure that the compiler does not adhere to the order (something in the back of my mind tells me that we enable some performance optimization flag that eliminates this behavior), I find you reasoning convincing 👍🏻

test/test_heuristic.cpp
burgholzer: Nah. It should be fine to keep it. One of the main difference between the two variants is heap (with unique_ptr) vs stack (with plain constructor) allocation. IIRC, it can actually become a problem to create too large objects in the stack (our DD package for example). Overall, I think unique_ptr is probably the better solution here.

test/test_heuristic.cpp
burgholzer: Seems fairly reasonable!

test/test_heuristic.cpp
burgholzer: Yeah, let's do that. I'll put it on my list of todos.

burgholzer:
> > One bigger thing: have you tested how much the information tracking/computation affects the performance? Considering runtime, but also memory. Would it make sense to "hide" these computation behind a "debug" or "monitoring" or "benchmarking" flag in the configuration and only track the metrics if the flag is enabled? Or is the overhead negligible anyway and we might as well include it all the time? I guess one of my main concern comes from the `Results` object being passed to Python which requires translating the C++ vector of results to Python. If the vector is large, this could be costly.
> 
> 
> 
> Oh, I did not even think about Python here. Using C++ directly, I did not see any significant change in mapping time (however, I also did not check memory usage, though I do not think there is any potential for a huge memory overhead). But I guess it cannot hurt to hide the feature behind a debugging flag in the configuration. Thank you for the suggestion!

I agree. It's probably not the worst thing (the data structure is rather tiny), but in order to maximize performance I believe it could be helpful to deactivate the metric tracking.

test/test_heuristic.cpp
EliasLF: Good idea. I did that now. Thank you!

burgholzer (approved):
Just pushed some commits with finishing touches here. 
Most importantly I took care of a case of undefined behavior (calling `back()` on an empty vector) and made the debug code in the mapper a little more compact/less spread out.
If CI checks turn all-green and you are ok with the changes I made, I'll proceed an merge.

EliasLF:
LGTM. Thank you for the refinements!

PR #282, opened 2023-03-30
⚡ Improve `SingleGateEncoder` Performance in Clifford Synthesis

pehamTom:
## Description

This PR improves the performance of the SingleGate encoding used in the gate optimization of the Clifford synthesis technique by reformulating the constraints in a simpler, more concise fashion. 

## Checklist:


src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
```diff
@@ -65,7 +101,7 @@ LogicTerm SingleGateEncoder::createSingleQubitGateConstraint(
             (tvars->r[pos + 1] ==
              (tvars->r[pos] ^ tvars->singleQubitRChange(pos, qubit, gate)));
 
-  return changes && createNoChange(pos, qubit, std::nullopt);
```
burgholzer, line 68: Is the function that this change belongs to even necessary any more after this PR?

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
```diff
@@ -119,6 +155,33 @@ LogicTerm SingleGateEncoder::createNoChange(
   return changes;
 }
 
+LogicTerm SingleGateEncoder::createNoChangeOnQubit(const std::size_t pos,
+                                                   const std::size_t q) {
+  auto noChange = LogicTerm(true);
+  noChange      = noChange && (tvars->x[pos + 1][q] == tvars->x[pos][q]);
+  noChange      = noChange && (tvars->z[pos + 1][q] == tvars->z[pos][q]);
+  return noChange;
```
burgholzer, line 163:
```suggestion
  return (tvars->x[pos + 1][q] == tvars->x[pos][q]) && (tvars->z[pos + 1][q] == tvars->z[pos][q]);
```
Should do the same thing and be much simpler, right?

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
```diff
@@ -119,6 +155,33 @@ LogicTerm SingleGateEncoder::createNoChange(
...
+}
+
+LogicTerm SingleGateEncoder::createNoGateOnQubit(const std::size_t pos,
+                                                 const std::size_t q) {
+  const auto& singleQubitGates = vars.gS[pos];
+  auto        noGate           = LogicTerm(true);
+  for (std::size_t i = 1; i < SINGLE_QUBIT_GATES.size(); ++i) {
+    noGate = noGate && !singleQubitGates[i][q];
+  }
```
burgholzer, line 172: This implicitly assumes that the `None` gate is at position 0. While this is a fair assumption, I believe it would be cleaner to 
 - either iterate over the SINGLE_QUBIT_GATES and skip over the None entry
 - get the index of the None entry before the loop and explicitly skip over its index in the loop.

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
```diff
@@ -119,6 +155,33 @@ LogicTerm SingleGateEncoder::createNoChange(
...
+  const auto& twoQubitGates = vars.gC[pos];
+  for (std::size_t i = 0; i < N; ++i) {
+    if (i == q) {
+      continue;
+    }
+    noGate = noGate && !twoQubitGates[i][q];
+    noGate = noGate && !twoQubitGates[q][i];
+  }
```
burgholzer, line 180: Probably stupid, but for larger qubit numbers, this is quite a long chain of ANDs. Would something like a commander encoding (i.e. some helper variables) be helpful here?
Or would you say that long chains of ANDs aren't really a problem?

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
```diff
@@ -119,6 +155,33 @@ LogicTerm SingleGateEncoder::createNoChange(
   return changes;
```
burgholzer, line 155: Is the `SingleGateEncoder::createNoChange(
    const std::size_t pos, const std::size_t except,
    const std::optional<std::size_t> except2)` function here still necessary after the changes here? 
Maybe phrased differently: I believe this function is still used for two-qubit gates. Does a change similar to the change for single-qubit gates also make sense there? 
Actually, it shouldn't be necessary there at all, right? the new constraints completely cover the NoGate changes.

burgholzer (changes requested):
Many many thanks for these changes! Really appreciate PRs like this. As often, they spark lots of ideas. You'll find in the comments below that most of the feedback is rather conceptual or trying to generalize concepts. Some of that might not make too much sense but I think some of the things would be really cool.

Nothing for this PR, but of course, it would also be interesting to adapt some of the changes here to the MultiGateEncoder where applicable. That should allow to also speed up the depth optimization.

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
```diff
@@ -36,16 +36,52 @@ void SingleGateEncoder::assertGateConstraints() {
...
+        LogicTerm::implies(singleQubitGates[gateToIndex((qc::OpType::H))][q],
+                           ((tvars->x[pos + 1][q] == tvars->z[pos][q]) &&
+                            (tvars->z[pos + 1][q] == tvars->x[pos][q]))));
+
+    // S and Sdag
+    lb->assertFormula(LogicTerm::implies(
+        (singleQubitGates[gateToIndex((qc::OpType::S))][q] ||
+         singleQubitGates[gateToIndex((qc::OpType::Sdag))][q]),
+        (tvars->z[pos + 1][q] == (tvars->z[pos][q] ^ tvars->x[pos][q]) &&
+         tvars->x[pos + 1][q] == tvars->x[pos][q])));
```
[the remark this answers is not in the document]
pehamTom: Okay, I was initially against this but I managed to find a somewhat elegant way of generalizing this that has 0 redundancy (elegant if you like `std::function` anyway). If the `singleQubit#Change` functions in `TableauEncoder` change, than the grouping that is done in the `SingleGateEncoder` should be adapted accordingly.

The question is whether this functions are even necessary in the first place exactly because we want to have the mapping exactly the other way around. Once I also adjust the `MultiGateEncoder` I think the functions will be no longer called from anywhere.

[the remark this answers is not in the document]
burgholzer: Looking forward to what you have came up with.
If some of the existing code can be removed/replaced, even better..

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
pehamTom: You are right. That was an oversight.

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
pehamTom: The function is not necessary anymore anyway. Removed it entirely.

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
pehamTom: Since the `None` entry is technically considered a gate one can actually include it in the loop anyway. But a `None` gate also doesn't hurt, so I will do it the first way you proposed.

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
pehamTom: In this particular case I would argue that it probably doesn't hurt too much. In my experience a commander encoding is useful for formulas that grow quadratically or worse. I'm not sure if grouping here would gain us much. 

I would say that this is something to benchmark for the future but not a pressing optimization.

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
pehamTom: It is not necessary anymore. The constraint that no gate should be applied to a qubit is independent of the specific gate anyway. So I think it makes more sense to formulate it independent of the single- and two-qubit constraints.

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
```diff
@@ -7,15 +7,18 @@
 
 #include "utils/logging.hpp"
 
+#include <functional>
+using namespace std::placeholders;
+
```
burgholzer, line 12:
```suggestion
```

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
```diff
@@ -36,36 +40,99 @@ void SingleGateEncoder::assertGateConstraints() {
...
+    DEBUG() << "Asserting gates on " << q;
+    assertZConstraints(pos, q);
+    assertXConstraints(pos, q);
+    assertRConstraints(pos, q);
+  }
+}
+
+std::vector<SingleGateEncoder::TransformationFamily>
+SingleGateEncoder::collectGateTransformations(
+    std::size_t pos, std::size_t qubit,
```
burgholzer, line 66:
```suggestion
    const std::size_t pos, const std::size_t qubit,
```

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
```diff
@@ -36,36 +40,99 @@ void SingleGateEncoder::assertGateConstraints() {
     TRACE() << "Asserting gate constraints at time " << t;
     assertSingleQubitGateConstraints(t);
     assertTwoQubitGateConstraints(t);
+    assertNoGateNoChangeConstraint(t);
+  }
+}
+
+void SingleGateEncoder::assertNoGateNoChangeConstraint(std::size_t pos) {
```
burgholzer, line 47:
```suggestion
void SingleGateEncoder::assertNoGateNoChangeConstraint(const std::size_t pos) {
```

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
```diff
@@ -36,36 +40,99 @@ void SingleGateEncoder::assertGateConstraints() {
...
+                                   std::vector<qc::OpType>{gate});
+    }
+  }
+  return transformations;
+}
 
-      lb->assertFormula(
-          LogicTerm::implies(singleQubitGates[gateToIndex(gate)][q], changes));
+void SingleGateEncoder::assertGatesImplyTransform(
+    std::size_t pos, std::size_t qubit,
```
burgholzer, line 87:
```suggestion
    const std::size_t pos, const std::size_t qubit,
```

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
```diff
@@ -36,36 +40,99 @@ void SingleGateEncoder::assertGateConstraints() {
...
+      gateOr = gateOr || singleQubitGates[gateToIndex(gate)][qubit];
     }
+    lb->assertFormula(LogicTerm::implies(gateOr, transformation));
   }
 }
 
-LogicTerm SingleGateEncoder::createSingleQubitGateConstraint(
-    const std::size_t pos, const std::size_t qubit, const qc::OpType gate) {
-  auto changes = LogicTerm(true);
+void SingleGateEncoder::assertZConstraints(std::size_t pos, std::size_t qubit) {
```
burgholzer, line 99:
```suggestion
void SingleGateEncoder::assertZConstraints(const std::size_t pos, const std::size_t qubit) {
```

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
```diff
@@ -36,36 +40,99 @@ void SingleGateEncoder::assertGateConstraints() {
...
+}
 
-  changes = changes && (tvars->x[pos + 1][qubit] ==
-                        tvars->singleQubitXChange(pos, qubit, gate));
-  changes = changes && (tvars->z[pos + 1][qubit] ==
-                        tvars->singleQubitZChange(pos, qubit, gate));
-  changes = changes &&
-            (tvars->r[pos + 1] ==
-             (tvars->r[pos] ^ tvars->singleQubitRChange(pos, qubit, gate)));
+void SingleGateEncoder::assertXConstraints(std::size_t pos, std::size_t qubit) {
```
burgholzer, line 112:
```suggestion
void SingleGateEncoder::assertXConstraints(const std::size_t pos, const std::size_t qubit) {
```

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
```diff
@@ -36,36 +40,99 @@ void SingleGateEncoder::assertGateConstraints() {
...
+  auto gateTransformations =
+      collectGateTransformations(pos, qubit, gatesToXTransformations);
+  for (auto& [transformation, _] : gateTransformations) {
+    transformation = tvars->x[pos + 1][qubit] == transformation;
+  }
+  assertGatesImplyTransform(pos, qubit, gateTransformations);
+}
 
-  return changes && createNoChange(pos, qubit, std::nullopt);
+void SingleGateEncoder::assertRConstraints(std::size_t pos, std::size_t qubit) {
```
burgholzer, line 125:
```suggestion
void SingleGateEncoder::assertRConstraints(const std::size_t pos, const std::size_t qubit) {
```

burgholzer (approved):
Many thanks. I'll quickly batch up those couple of missing `const`s and the superfluous header and then proceed to merge.

PR #284, opened 2023-04-04
🚸 Allow for setting minimal timesteps for binary search.

pehamTom:
## Description

This PR adds a setting to the config that allows to set the lower limit for the binary search.

Sometimes a lower bound for the circuit might already be known and one can save on some calls to the Solver by limiting the search space during binary search.

## Checklist:

### DIFF
--- a/include/cliffordsynthesis/Configuration.hpp
+++ b/include/cliffordsynthesis/Configuration.hpp
@@ -16,6 +16,7 @@ struct Configuration {
 
   /// General configuration for the synthesis algorithm
   std::size_t    initialTimestepLimit    = 0U;
+  std::size_t    minimalTimeSteps        = 0U;
   bool           useMaxSAT               = false;
   TargetMetric   target                  = TargetMetric::Gates;
   bool           useSymmetryBreaking     = true;
@@ -37,6 +38,7 @@ struct Configuration {
   [[nodiscard]] nlohmann::json json() const {
     nlohmann::json j;
     j["initial_timestep_limit"] = initialTimestepLimit;
+    j["minimal_timesteps"]      = minimalTimeSteps;
     j["use_max_sat"]            = useMaxSAT;
     j["target_metric"]          = toString(target);
     j["use_symmetry_breaking"]  = useSymmetryBreaking;
--- a/include/cliffordsynthesis/Configuration.hpp
+++ b/include/cliffordsynthesis/Configuration.hpp
@@ -16,7 +16,7 @@ struct Configuration {
 
   /// General configuration for the synthesis algorithm
   std::size_t    initialTimestepLimit    = 0U;
-  std::size_t    minimalTimeSteps        = 0U;
+  std::size_t    minimalTimesteps        = 0U;
   bool           useMaxSAT               = false;
   TargetMetric   target                  = TargetMetric::Gates;
   bool           useSymmetryBreaking     = true;
@@ -38,7 +38,7 @@ struct Configuration {
   [[nodiscard]] nlohmann::json json() const {
     nlohmann::json j;
     j["initial_timestep_limit"] = initialTimestepLimit;
-    j["minimal_timesteps"]      = minimalTimeSteps;
+    j["minimal_timesteps"]      = minimalTimesteps;
     j["use_max_sat"]            = useMaxSAT;
     j["target_metric"]          = toString(target);
     j["use_symmetry_breaking"]  = useSymmetryBreaking;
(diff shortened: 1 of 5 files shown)


src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -48,8 +48,11 @@ void CliffordSynthesizer::synthesize(const Configuration& config) {
   // SAT problem repeatedly with increasing timestep limits until a satisfying
   // assignment is found. This uses the general SAT encoding without any
   // objective function regardless of the configuration.
-  const auto [lower, upper] = determineUpperBound(encoderConfig);
+  auto [lower, upper] = determineUpperBound(encoderConfig);
 
+  if (lower == 0) {
+    lower = config.minimalTimeSteps;
+  }
```
burgholzer, line 55: I am not quite sure I agree with this change.
Shouldn't this be something like
```c++
if (config.minimalTimeSteps > 0) {
  lower = std::max(lower, config.minimalTimeSteps);
}
```
which would match the description of the parameter a lot more.

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -66,7 +69,7 @@ void CliffordSynthesizer::synthesize(const Configuration& config) {
     gateOptimalSynthesis(encoderConfig, lower, upper);
     break;
   case TargetMetric::Depth:
-    depthOptimalSynthesis(encoderConfig, lower, upper);
+    depthOptimalSynthesis(encoderConfig, config.minimalTimeSteps, upper);
```
burgholzer, line 72: This should still be passed `lower`, right?

include/cliffordsynthesis/Configuration.hpp
```diff
@@ -16,6 +16,7 @@ struct Configuration {
 
   /// General configuration for the synthesis algorithm
   std::size_t    initialTimestepLimit    = 0U;
+  std::size_t    minimalTimeSteps        = 0U;
```
burgholzer, line 19:
```suggestion
  std::size_t    minimalTimesteps        = 0U;
```
For consistency. Otherwise you would have to change the other occurrences to `minimal_time_steps`.
Another possibility: name this `minimalTimestepLimit` in accordance with the parameter right above it.

burgholzer (changes requested):
Many thanks. Please see below for a couple of requests/comments that I think need addressing before this can be merged.

src/cliffordsynthesis/CliffordSynthesizer.cpp
burgholzer: Now that I think about it: shouldn't this actually come before the `determineUpperBound` call?
That might actually safe some SAT solver calls there.

pehamTom: I added an option to the `EncoderConfig` so the information is available in `determineUpperBound` without changing the funtion signature.

src/cliffordsynthesis/CliffordSynthesizer.cpp
pehamTom: Yes of course.

include/cliffordsynthesis/Configuration.hpp
pehamTom: Yeah that's a typo.

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -122,7 +121,7 @@ CliffordSynthesizer::determineUpperBound(EncoderConfig config) {
     return {0U, config.timestepLimit};
   }
 
-  std::size_t lowerBound = 0U;
+  std::size_t lowerBound = config.minimalTimesteps;
   std::size_t upperBound = config.timestepLimit;
```
burgholzer, line 125: I believe a couple of things are still not right here:
 - At the top, in case a circuit is provided, the result should most likely be `{config.minimalTimesteps, config.timestepLimit}`.
 - This leads to another point that also affects the main iteration: it should be ensured that `upperBound` is at least as large as `lowerBound`. Otherwise this might produce nonsense.

I'd say it's propbably best to move such a sanity check in front of the `if(results.sat())` check, i.e., something like:
```c++
config.timestepLimit = std::max(config.timestepLimit, config.minimalTimesteps);
std::size_t lowerBound = config.minimalTimesteps;
std::size_t upperBound = config.timestepLimit;

if (results.sat()) {
  return {lower, upper};
}
```

burgholzer (changes requested):

src/cliffordsynthesis/CliffordSynthesizer.cpp
burgholzer: No worries. That's what these PR reviews are for.

PR #304, opened 2023-05-08
Improvements for the Paper "Depth-Optimal Synthesis of Clifford Circuits with SAT Solvers"

pehamTom:
## Description

This PR introduces a couple of changes simultaneously that were introduced during the work culminating in [this paper](https://arxiv.org/abs/2305.01674). These are:

- A parallel heuristic based on vertical circuit decomposition.
- Various improvements to the symmetry breaking constraints
- Allowing for setting of solver parameters from the Python side
- Some smaller fixes and improvements

Fixes # (issue)

## Checklist:


include/cliffordsynthesis/encoding/GateEncoder.hpp
```diff
@@ -55,10 +55,11 @@ class GateEncoder {
   void extractCircuitFromModel(Results& res, logicbase::Model& model);
 
   [[nodiscard]] auto* getVariables() { return &vars; }
-
+  //
```
burgholzer, line 58:
```suggestion
 
```
seems like an unnecessary change to me

mqt/qmap/pyqmap.pyi
burgholzer, line 1: This file is missing some of the new options for the synthesis

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -189,10 +208,12 @@ void CliffordSynthesizer::depthOptimalSynthesis(
...
     // circuit.
     runMaxSAT(config);
+  } else if (configuration.linearSearch) {
+    runLinearSearch(config.timestepLimit, lower, upper, config);
   } else {
     // The binary search approach calls the SAT solver repeatedly with varying
-    // timestep (=depth) limits T until a solution with depth T is found, but no
-    // solution with depth T-1 could be determined.
+    // timestep (=gate) limits T until a solution with T gates is found, but no
+    // solution with T-1 gates could be determined.
```
burgholzer, line 216: this change seems wrong to me.

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -395,4 +416,82 @@ void CliffordSynthesizer::updateResults(const Configuration& config,
...
 }
+
+// assume canonical sorting of gates
+std::vector<std::size_t> getLayers(const qc::QuantumComputation& qc) {
+  std::vector<std::size_t> layerNum{};
+  layerNum.resize(qc.size());
+  std::vector<std::size_t> layers{};
+  std::size_t              layer = 0U;
+  for (std::size_t i = 0; i < qc.size(); ++i) {
+    const auto& gate = *(qc.begin() + i);
```
burgholzer, line 427: Maybe use `qc.at(i)` instead?

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -395,4 +416,82 @@ void CliffordSynthesizer::updateResults(const Configuration& config,
...
+void CliffordSynthesizer::depthHeuristicSynthesis() {
+  INFO() << "Optimizing Circuit with Heuristic" << std::endl;
+  auto optimalConfig                 = configuration;
+  optimalConfig.heuristic            = false;
+  optimalConfig.target               = TargetMetric::Depth;
+  optimalConfig.initialTimestepLimit = configuration.splitSize;
+
+  qc::CircuitOptimizer::reorderOperations(*initialCircuit);
+  qc::QuantumComputation   optCircuit{initialCircuit->getNqubits()};
+  std::vector<std::size_t> layers = getLayers(*initialCircuit);
```
burgholzer, line 455: Maybe I have overlooked something, but I believe this can be `const`

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -395,4 +416,82 @@ void CliffordSynthesizer::updateResults(const Configuration& config,
...
+
+  std::vector<std::future<std::shared_ptr<qc::QuantumComputation>>> subCircuits;
+  for (std::size_t i = 0; i < layers.size() - 1; i += configuration.splitSize) {
+    std::size_t const          startIdx = layers[i];
+    std::optional<std::size_t> endIdx;
```
burgholzer, line 460: Why is this an optional, when the value is always set?

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -395,4 +416,82 @@ void CliffordSynthesizer::updateResults(const Configuration& config,
...
+              initialCircuit, startIdx, endIdx.value(), optimalConfig);
+        }));
+  }
+
+  for (auto& subCircuit : subCircuits) {
+    const auto& circ = subCircuit.get();
+    for (const auto& op : *circ) {
+      optCircuit.emplace_back(op->clone());
+    }
+  }
```
burgholzer, line 482: This seems a bit wasteful to me as it incurs real copies of all operations. The sub circuits are not needed any more, so their operations can simply be moved to the optimal circuit without incurring copies.

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
```diff
@@ -92,11 +100,15 @@ LogicTerm encoding::MultiGateEncoder::createTwoQubitGateConstraint(
...
-
+  const auto& newRChanges = LogicTerm::ite(
+      vars.gC[pos][ctrl][trgt], tvars->twoQubitRChange(pos, ctrl, trgt),
+      LogicTerm(0, static_cast<std::int16_t>(S)));
+  splitXorR(newRChanges, pos);
+  //  rChanges =
+  //      rChanges ^ LogicTerm::ite(vars.gC[pos][ctrl][trgt],
+  //                                tvars->twoQubitRChange(pos, ctrl, trgt),
+  //                                LogicTerm(0, static_cast<std::int16_t>(S)));
+  //
```
burgholzer, line 111: commented out code can be removed
```suggestion
```

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
```diff
@@ -58,10 +60,16 @@ void encoding::MultiGateEncoder::assertSingleQubitGateConstraints(
...
+    const auto& change =
+        LogicTerm::ite(vars.gS[pos][gateToIndex(gate)][qubit],
+                       tvars->singleQubitRChange(pos, qubit, gate),
+                       LogicTerm(0, static_cast<std::int16_t>(S)));
+    splitXorR(change, pos);
+    //    rChanges =
+    //        rChanges ^ LogicTerm::ite(vars.gS[pos][gateToIndex(gate)][qubit],
+    //                                  tvars->singleQubitRChange(pos, qubit,
+    //                                  gate), LogicTerm(0,
+    //                                  static_cast<std::int16_t>(S)));
```
burgholzer, line 72: commented out code can be removed
```suggestion
```

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
```diff
@@ -132,19 +144,73 @@ void MultiGateEncoder::assertTwoQubitGateOrderConstraints(
...
+      LogicTerm::implies(vars.gC[pos][ctrl][trgt], !gSNext[zIndex][ctrl]));
+  lb->assertFormula(
+      LogicTerm::implies(vars.gC[pos][trgt][ctrl], !gSNext[zIndex][trgt]));
+
+  constexpr auto sIndex = gateToIndex(qc::OpType::S);
+  lb->assertFormula(
+      LogicTerm::implies(vars.gC[pos][ctrl][trgt], !gSNext[sIndex][ctrl]));
+  lb->assertFormula(
+      LogicTerm::implies(vars.gC[pos][trgt][ctrl], !gSNext[sIndex][trgt]));
+}
```
burgholzer, line 198: This part is missing some comments explaining the symmetry breaker. For example, I am not 100% convinced, that the last implication is true or at least guaranteed to preserve depth-optimality.

Furthermore, I think it is missing rules for the Y gate and the Sdag gate (if the S rule really holds), which should be analogous to the above rules.

Last but not least, many of these constraints can actually be combined so that multiple implications can be reduced to a single one. (`LogicTerm::implies(vars.gC[pos][ctrl][trgt], X ... Y ... Z)`)

src/cliffordsynthesis/encoding/SATEncoder.cpp
```diff
@@ -130,7 +141,6 @@ Results SATEncoder::run() {
   const auto start = std::chrono::high_resolution_clock::now();
 
   createFormulation();
-  produceInstance();
```
burgholzer, line 133: has this been removed on purpose?

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
burgholzer, line 1: Should some of the more advanced symmetry breakers also be added here? Phrased differently: are we missing something here that makes sense to be added?

test/cliffordsynthesis/test_synthesis.cpp
burgholzer, line 1: Could we add at least a couple of slightly more sophisticated tests for the heuristic part?
Could we also add such tests on the Python side to make sure people can actually use this?
Could we add a test that ensures that the results with Symmetry Breaking are the same as without using it? Would be an important verification step to make sure we don't introduce any optimizations that harm optimality.

burgholzer (changes requested):
Just managed to quickly go through everything here. Many thanks for all the work.
You can find most of my suggestions inline as comments. Should not take too much time to address them.
A couple more things:
- Extending the documentation a little would be great to highlight the new options as well as to add the preprint
- Already annotated in the comments, but still important: more elaborate unit tests would really help to make sure everything is working and keeps working in the future.
- Please choose a proper PR title ;o)

src/cliffordsynthesis/CliffordSynthesizer.cpp
pehamTom: Not entirely sure what I was thinking there...

src/cliffordsynthesis/CliffordSynthesizer.cpp
pehamTom: Done.

src/cliffordsynthesis/CliffordSynthesizer.cpp
pehamTom: Weird that not even SonarLint suggested that.

src/cliffordsynthesis/CliffordSynthesizer.cpp
pehamTom: I guess it is a habit. I tend to use optionals whenever a variable isn't directly initialized. But in this case this should never be a problem.

src/cliffordsynthesis/CliffordSynthesizer.cpp
burgholzer: That's what the `const` police (aka me) is for 😜

src/cliffordsynthesis/CliffordSynthesizer.cpp
burgholzer: I believe that's a reasonable habit. In cases like these, where the initialization happens right below and there is never a case where the variable might not be initialized (you didn't call `has_value()` once), it might be superfluous.

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
pehamTom: Done

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
pehamTom: Done

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
pehamTom: Very good catch. The rules should be all correct. Any Z-rotation commutes with the control a CNOT. But this might indeed break depth-optimality. Only the symmetry where there is a gate after both the control and target are guaranteed to preserve depth-optimality because no gates can be parallel after the CNOTs. I removed all those symmetry breakers.

On the other hand I included symmetry breakers for the case when two Z or X gates are **before** the CNot, because these can also just be replaced by a single gate after the CNOT without changing the depth.

src/cliffordsynthesis/encoding/SATEncoder.cpp
pehamTom: Yes. `produceInstance` is invoked in the call to `solve` in the LogicBlocks library. Even worse, there was no check whether the instance was already created, basically doubling the size of the encoding without any benefit.

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
burgholzer: Yeah, exactly. With two-qubit gates, one has to always consider both qubits either before or after the two-qubit gate.
Is there a general rule of the form:
```
-- [I | X | Y | Z | H] -- q0 --
-- [I | X | Y | Z | H] -- q1 --

==

-- [q0 | q1] -- [I | X | Y | Z | H] --
-- [q1 | q0] -- [I | X | Y | Z | H] --
```
?

I am not sure how large the set of single qubit gates can really be. Can it also include the S and Sdag gate? Is no such general statement possible at all? What is the most general two-qubit "normal form"?

src/cliffordsynthesis/encoding/SATEncoder.cpp
burgholzer: Oh wow. What a waste.

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
pehamTom: The S gate cannot be commuted through the target qubit of the CX gate. If that were the case then we could push all single-qubit gates to the end of the circuit. This can only be done with Paulis however. So the current rules should be the most general.

mqt/qmap/pyqmap.pyi
pehamTom: Done

src/cliffordsynthesis/encoding/SingleGateEncoder.cpp
pehamTom: The two-qubit constraints could be added. But since there is only one gate per timestep the constraints would have to be formulated over 5 timesteps. I guess we could add them. I assume we don't need to consider all permutations of the single-qubit gates since there is an order already imposed on these gates?

burgholzer: Hm. That sounds a bit cumbersome. In that case, and to safe some effort, it's probably best to leave this as-is.

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
```diff
@@ -132,19 +134,62 @@ void MultiGateEncoder::assertTwoQubitGateOrderConstraints(
...
+  const auto cnotBefore  = vars.gC[pos][ctrl][trgt] || vars.gC[pos][trgt][ctrl];
+  lb->assertFormula(LogicTerm::implies(cnotBefore, !zzAfter && !xxAfter));
+
+  // No Zs or Xs on both qubits before CNOT
+  const auto xxBefore = gSNow[xIndex][ctrl] && gSNow[xIndex][trgt];
+  const auto zzBefore = gSNow[zIndex][ctrl] && gSNow[zIndex][trgt];
+  const auto cnotAfter =
+      vars.gC[pos + 1][ctrl][trgt] || vars.gC[pos + 1][trgt][ctrl];
+  lb->assertFormula(LogicTerm::implies(cnotAfter, !zzBefore && !xxBefore));
+}
```
burgholzer, line 177: Maybe I just don't quite get it, but aren't these kind of incomplete and inefficient?
Doesn't some variation of this rule hold for any combination of Paulis before or after a CNOT? At the very least there should also be a rule for double Y gates. But, as said, I believe the rule can be more general than that.
Furthermore, the number of constraints could be reduced by combining the before and after rules (plus some special handling for the corner cases).

test/cliffordsynthesis/test_synthesis.cpp
```diff
@@ -319,4 +349,66 @@ TEST_P(SynthesisTest, TestDestabilizerTwoQubitGates) {
...
+  qc.h(1);
+  config.heuristic = true;
+  config.splitSize = 2;
+  config.target    = TargetMetric::Depth;
+  auto synth       = CliffordSynthesizer(qc);
+  synth.synthesize(config);
+  EXPECT_EQ(synth.getResults().getDepth(), 3);
+}
+
+using namespace qc::literals;
```
burgholzer, line 399: Instead of repeating the using directive here twice, consider moving it to the very top (only once 🙃)

test/python/test_cliffordsynthesis.py
```diff
@@ -115,6 +116,30 @@ def test_optimize_clifford_gates_at_minimal_two_qubit_gates(test_config: Configu
     print("\n", circ)
 
 
+@pytest.mark.parametrize("test_config", create_circuit_tests())
+def test_heuristic(test_config: Configuration) -> None:
+    """Test heuristic synthesis method."""
+    print(test_config.initial_circuit)
+    sys.stdout.flush()
```
burgholzer, line 123: This, and the `sys` import seem like a leftover from debugging. Consider removing them. Same goes for the flush down below.

test/python/test_cliffordsynthesis.py
```diff
@@ -115,6 +116,30 @@ def test_optimize_clifford_gates_at_minimal_two_qubit_gates(test_config: Configu
...
+    circ, results = qmap.optimize_clifford(
```
burgholzer, line 124: Both result and the other result below are unused and should be replaced by `_`

docs/source/Synthesis.ipynb
```diff
@@ -191,7 +191,75 @@
...
+    "qc_opt, results = qmap.optimize_clifford(circuit=qc, heuristic=False, include_destabilizers=True, target_metric=\"depth\")\n",
+    "\n",
+    "qc_opt.draw(output=\"mpl\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "However, the heuristic still gives a good depth reduction in many cases."
```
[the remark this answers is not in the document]
pehamTom: For such a small example not really. I could use a larger one but I just wanted to demonstrate how to use it and what output to expect.

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -395,4 +417,100 @@ void CliffordSynthesizer::updateResults(const Configuration& config,
...
+  for (const auto& qubit : gate.getUsedQubits()) {
+    layerNum[qubit] = layer;
+  }
+  ++i;
+}
+
+// assume canonical sorting of gates
+std::vector<std::size_t> getLayers(const qc::QuantumComputation& qc) {
+  std::vector<std::size_t> layerNum{};
+  layerNum.resize(qc.size());
```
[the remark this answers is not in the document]
pehamTom: I would be careful here. Since it's a vector of `std::size_t`, passing a size to the constructor creates a vector with one element instead of a vector with a pre-defined size. So this is the correct way here.

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -395,4 +417,100 @@ void CliffordSynthesizer::updateResults(const Configuration& config,
...
+                 std::vector<std::size_t>& layers,
+                 std::vector<std::size_t>& layerNum, std::size_t& layer) {
+  for (const auto& qubit : gate.getUsedQubits()) {
+    if (layerNum[qubit] >= layer) {
+      ++layer;
+      layers.emplace_back(i);
+      break;
+    }
+  }
+  for (const auto& qubit : gate.getUsedQubits()) {
```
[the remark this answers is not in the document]
pehamTom: Good point

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -395,4 +417,100 @@ void CliffordSynthesizer::updateResults(const Configuration& config,
...
+        gateToLayer(*subGate, i, layers, layerNum, layer);
+      }
+    } else {
+      gateToLayer(*gate, i, layers, layerNum, layer);
+    }
+  }
+
+  if (layers.back() < qc.getNindividualOps()) {
+    layers.emplace_back(qc.getNindividualOps());
+  }
```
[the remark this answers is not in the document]
pehamTom: Changed it

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -395,4 +417,100 @@ void CliffordSynthesizer::updateResults(const Configuration& config,
...
+  return layers;
+}
+
+void CliffordSynthesizer::depthHeuristicSynthesis() {
+  INFO() << "Optimizing Circuit with Heuristic" << std::endl;
```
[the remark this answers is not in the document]
pehamTom: Removed it.

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -395,4 +417,100 @@ void CliffordSynthesizer::updateResults(const Configuration& config,
...
+  auto optimalConfig = configuration;
```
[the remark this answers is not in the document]
pehamTom: Very well :)

docs/source/Synthesis.ipynb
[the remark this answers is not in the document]
burgholzer: Alright. In that case, just leave it.

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
```diff
@@ -132,19 +134,62 @@ void MultiGateEncoder::assertTwoQubitGateOrderConstraints(
...
+  lb->assertFormula(
+      LogicTerm::implies(vars.gC[pos][ctrl][trgt], !gCNext[ctrl][trgt]));
+  lb->assertFormula(
+      LogicTerm::implies(vars.gC[pos][trgt][ctrl], !gCNext[trgt][ctrl]));
 
   // if no gate is applied to both qubits, no CNOT on them can be applied in the
   // next time step.
-  const auto noneIndex     = gateToIndex(qc::OpType::None);
-  const auto noGate        = gSNow[noneIndex][ctrl] && gSNow[noneIndex][trgt];
+  // no gate on both qubits => no CNOT on them in the next time step.
```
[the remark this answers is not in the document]
pehamTom: Removed duplicate

test/cliffordsynthesis/test_synthesis.cpp
pehamTom: C&P mistake. Fixed it.

src/cliffordsynthesis/CliffordSynthesizer.cpp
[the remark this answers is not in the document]
burgholzer: Right. Nevermind then 😌

test/python/test_cliffordsynthesis.py
pehamTom: True. Removed.

test/python/test_cliffordsynthesis.py
pehamTom: Done.

docs/source/Synthesis.ipynb
```diff
@@ -228,9 +296,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "general",
    "language": "python",
-   "name": "python3"
+   "name": "general"
```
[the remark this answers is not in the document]
pehamTom: Changed my kernel locally and forgot that that writes into the ipynb file

docs/source/refs.bib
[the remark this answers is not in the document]
pehamTom: Done

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
```diff
@@ -132,19 +134,61 @@ void MultiGateEncoder::assertTwoQubitGateOrderConstraints(
...
+  constexpr auto hIndex = gateToIndex(qc::OpType::H);
+  const auto     hh     = gSNow[hIndex][ctrl] && gSNow[hIndex][trgt];
 
-  // if no gate is applied to both qubits, no CNOT on them can be applied in the
-  // next time step.
-  const auto noneIndex     = gateToIndex(qc::OpType::None);
-  const auto noGate        = gSNow[noneIndex][ctrl] && gSNow[noneIndex][trgt];
   const auto noFurtherCnot = !gCNext[ctrl][trgt] && !gCNext[trgt][ctrl];
-  lb->assertFormula(LogicTerm::implies(noGate, noFurtherCnot));
+  lb->assertFormula(LogicTerm::implies(noGate || hh, noFurtherCnot));
```
burgholzer, line 158: I really hate to bother you again with something. But the realization here still is not perfect.
The H constraints are now included twice.
The "proper" left-hand side for the implication would be `noGateOnBoth || ((X || Y || Z || H on control) && (X || Y || Z || H on target)`

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
```diff
@@ -132,19 +134,61 @@ void MultiGateEncoder::assertTwoQubitGateOrderConstraints(
...
+
+  // No Combination of Paulis or Hadamards on both Qubits before a CNOT
+  // These gates can be just pushed through to the other side
+
+  if (pos == 0) { // In the beginning parallel gates are allowed
+    return;
+  }
```
burgholzer, line 165: I am fairly sure that this condition is not necessary. Why shouldn't these rules be allowed in the very beginning. The scenario in the beginning is just the same as at any other point in the circuit.
A block of two single-qubit gates (paulis+H) and a CNOT can _always_ be replaced by a CNOT and at most two single-qubit gates.
I could image what you were seeing was a consequence of the duplicate rules highlighted in the other comment.

burgholzer (changes requested):
Almost there 🙃

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
pehamTom: I even thought about it and forgot it again. Good catch

src/cliffordsynthesis/encoding/MultiGateEncoder.cpp
pehamTom: Only parallel H and Paulis are a problem so I added a special case for the first timestep that encodes only the other combinations

pehamTom: No but you could push the Pauli from the end to the beginning to reduce the depth by 1. But the symmetry breaker wouldn't allow that. That's why we have to exclude the first layer.

burgholzer (approved):
Many thanks. This looks great now 🎉

PR #316, opened 2023-05-23
Fix cnot reversal in heuristic

EliasLF:
## Description

The current heuristic implementation, when determining if a reversal will be necessary for some given CNOT on a unidirectional architecture, considers only the 2 edges next to the 2 qubits involved in the CNOT along the shortest path (i.e. it assumes that one of the 2 qubits will eventually move all the way next to current position of the other qubit, instead of possibly meeting somewhere in the middle). However, if there is _any_ edge along the shortest path, a reversal might in fact not be necessary. Thereby, the current implementation overestimates the true cost (i.e. violates admissibility) in some cases.

This PR fixes this by only applying the reversal cost if no edge along the shortest path is oriented in the right direction.

Additionally, a minor bug fix in benchmarking: The checks for `effectiveBranchingFactor > averageBranchingFactor` were removed, as I realized that this can actually occur for heavily unbalanced search trees (commonly occuring for very flat heuristics).

## Checklist:

### DIFF
--- a/include/Architecture.hpp
+++ b/include/Architecture.hpp
@@ -364,20 +364,6 @@ protected:
   void createDistanceTable();
   void createFidelityTable();
 
-  static double dijkstraNodeToCost(const Dijkstra::Node& node) {
-    // Dijkstra determines the minimal path cost from one physical qubit to
-    // another.  In the non-fidelity case we are only interested in swapping 2
-    // logical qubits next to each other. Therefore the last swap cost has to
-    // be ignored. That cost is stored in the field `node.prevCost` of each
-    // node.
-    if (node.containsCorrectEdge) {
-      return node.prevCost;
-    }
-    // in case the last edge is a back-edge, we will need to reverse the CNOT,
-    // executed on that edge
-    return node.prevCost + COST_DIRECTION_REVERSE;
-  }
-
   // added for teleportation
   static bool contains(const std::vector<int>& v, const int e) {
     return std::find(v.begin(), v.end(), e) != v.end();
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -9,7 +9,8 @@
 
 void Dijkstra::buildTable(const std::uint16_t n, const CouplingMap& couplingMap,
                           Matrix& distanceTable, const Matrix& edgeWeights,
-                          const double reversalCost, const bool removeLastEdge) {
+                          const double reversalCost,
+                          const bool   removeLastEdge) {
   distanceTable.clear();
   distanceTable.resize(n, std::vector<double>(n, -1.));
 
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -23,7 +23,10 @@ void Dijkstra::buildTable(const std::uint16_t n, const CouplingMap& couplingMap,
       nodes.at(j).cost                = -1.;
       nodes.at(j).prevCost            = -1.;
     }
-
+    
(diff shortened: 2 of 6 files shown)


include/Architecture.hpp
```diff
@@ -364,20 +364,6 @@ class Architecture {
...
-    // Dijkstra determines the minimal path cost from one physical qubit to
-    // another.  In the non-fidelity case we are only interested in swapping 2
-    // logical qubits next to each other. Therefore the last swap cost has to
-    // be ignored. That cost is stored in the field `node.prevCost` of each
-    // node.
-    if (node.containsCorrectEdge) {
-      return node.prevCost;
-    }
-    // in case the last edge is a back-edge, we will need to reverse the CNOT,
-    // executed on that edge
```
burgholzer, line 377: Can we add comments along those lines to the newly introduced parts in the code so it becomes clearer what is going on? (E.g. why costs are no longer initialized with 0, but cost-reversal)

src/utils.cpp
```diff
@@ -68,6 +73,10 @@ void Dijkstra::dijkstra(const CouplingMap& couplingMap,
         newNode.prevCost = current->cost;
         newNode.pos      = to;
         newNode.containsCorrectEdge = correctEdge;
+        if (newNode.containsCorrectEdge && !current->containsCorrectEdge) {
+          newNode.cost -= reversalCost;
+          newNode.prevCost -= reversalCost;
+        }
```
burgholzer, line 79: I might be wrong here, but can't it happen that this condition triggers multiple times?
For example on a path with 
`n0` -- correct -- `n1` -- wrong -- `n2` -- correct -- `n3` the condition would be triggered twice, right?

test/test_general.cpp
burgholzer, line 1: Can we maybe add a regression test that makes sure that the proper cost is now returned for an example that previously did not return the correct result?

burgholzer (changes requested):
Thanks for identifying the problem and creating a fix. I just have two requests/questions, see below.

src/utils.cpp
EliasLF: I am quite certain that this condition only fires at most once per path, as any new node is always initialized with the same `containsCorrectEdge` as the previous node and only flipped to `true` given the right circumstances (and never back to `false`). In a way `containsCorrectEdge` encodes if the path up to and including this edge contains a forward-edge, instead of only the current edge. Therefore the flip from `false` to `true` only happens once.

In your example the condition would be triggered in the first edge and after that both `newNode.containsCorrectEdge` and `current->containsCorrectEdge` will both always be true (as `newNode.containsCorrectEdge` is initialized from `current->containsCorrectEdge` and can only be changed to true)

However, I see, how this might be unintuitive, and will add some comments explaining all of it as suggested by you.

burgholzer: Ahhh 😌 now that you say it, I also see that. Maybe it was the GitHub diff that increased the confusion because I kept seeing the `false` initialization, but of course that was a deletion.
Then I am totally convinced that this is correct.
Adding a comment sounds like a good idea though!

include/Architecture.hpp
EliasLF: I added some comments and doc strings that should now cover everything, I think.

test/test_general.cpp
EliasLF: I added such a test just now

burgholzer (approved):
Many thanks. LGTM 👍🏻

PR #342, opened 2023-07-08
Feature/TwoQubitEncoding

avelsh:
Added functionality for SQG-TQG shape encoding


src/cliffordsynthesis/encoding/SATEncoder.cpp
```diff
@@ -50,10 +51,14 @@ void SATEncoder::createFormulation() {
   const auto start = std::chrono::high_resolution_clock::now();
   initializeSolver();
 
-  const std::size_t s = config.targetTableau->hasDestabilizers() &&
-                                config.initialTableau->hasDestabilizers()
-                            ? 2U * N
-                            : N;
+  std::size_t s = config.targetTableau->hasDestabilizers() &&
```
pehamTom, line 54: Is there a reason that this is not `const` anymore?

src/cliffordsynthesis/encoding/STQGatesEncoder.cpp
```diff
@@ -0,0 +1,236 @@
...
+        const std::string gName = "g_" + std::to_string(t) + "_" +
+                                  toString(gate) + "_" + std::to_string(q);
+        TRACE() << "Creating variable " << gName;
+        g.emplace_back(lb->makeVariable(gName));
+      }
+    }
+  }
+}
+
+void encoding::STQGatesEncoder::createTwoQubitGateVariables() {
```
pehamTom, line 33: As it stands, much code is repeated from `GateEncoder.cpp`. You could call the base class method and then loop over the timesteps to insert the identity variables separately. Or don't overwrite the method and use `g_i_i` as the variable for the identity gate.

src/cliffordsynthesis/encoding/STQGatesEncoder.cpp
```diff
@@ -0,0 +1,236 @@
...
+}
+
+void encoding::STQGatesEncoder::collectTwoQubitGateVariables(
+    const std::size_t pos, const std::size_t qubit, const bool target,
+    LogicVector& variables) const {
+  const auto& twoQubitGates = vars.gC[pos];
+  const auto  n             = twoQubitGates.size();
+  for (std::size_t q = 0; q < n; ++q) {
+    if (q == qubit) {
+      if (N > variables.size()) {
```
pehamTom, line 67: I think I don't get this check. You could check for the `target` flag. This way you would also only add the variable for the identity once.

src/cliffordsynthesis/encoding/STQGatesEncoder.cpp
```diff
@@ -0,0 +1,236 @@
...
+    }
+  }
+}
+
+void encoding::STQGatesEncoder::assertGateConstraints() {
+  DEBUG() << "Asserting gate constraints";
+  for (std::size_t t = 0U; t < T; ++t) {
+    TRACE() << "Asserting gate constraints at time " << t;
+
+    const std::size_t pos = t < T / 2 ? t : t % (T / 2);
```
pehamTom, line 117: As the Linter says, this is a possible division by Zero if `T` is 0. Also, I think it would improve clarity to add a class member that is essentially just `T/2` that is set in the constructor. Then `T` would be the number of tableaus and `T/2` would be the number of layers.

src/cliffordsynthesis/encoding/STQGatesEncoder.cpp
```diff
@@ -0,0 +1,236 @@
...
+  }
+}
+
+void encoding::STQGatesEncoder::assertTwoQubitGateConstraints(
+    const std::size_t pos) {
+  const auto& twoQubitGates = vars.gC[pos];
+  for (std::size_t ctrl = 0U; ctrl < N; ++ctrl) {
+    for (std::size_t trgt = 0U; trgt < N; ++trgt) {
+      if (ctrl == trgt) {
+        continue;
```
pehamTom, line 151: There is a constraint that needs to be added here. You also have to assert that when the `None` gate is set in the two-qubit layer there are no changes. At the moment, when the solver decides to set the identity the tableau variables are unconstrained and can be anything.

src/cliffordsynthesis/encoding/STQGatesEncoder.cpp
```diff
@@ -0,0 +1,236 @@
...
+      }
+      const auto changes = createTwoQubitGateConstraint(pos, ctrl, trgt);
+      lb->assertFormula(LogicTerm::implies(twoQubitGates[ctrl][trgt], changes));
+
+      DEBUG() << "Asserting CNOT on " << ctrl << " and " << trgt;
+    }
+  }
+}
+
+LogicTerm encoding::STQGatesEncoder::createTwoQubitGateConstraint(
```
pehamTom, line 161: If I'm not missing something here, this method can be removed as it is just a duplicated from the base class.

include/cliffordsynthesis/encoding/STQGatesEncoder.hpp
```diff
@@ -0,0 +1,58 @@
...
+                                    bool                    target,
+                                    logicbase::LogicVector& variables) const;
+
+  // TODO: remove later
+  [[nodiscard]] std::vector<TransformationFamily> collectGateTransformations(
+      std::size_t pos, std::size_t qubit,
+      const GateToTransformation& gateToTransformation) override;
+
+  void
+  extractSingleQubitGatesFromModel(std::size_t pos, logicbase::Model& model,
```
pehamTom, line 53: You need to overwrite the `GateEncoder::extractCircuitFromModel` method as well. At the moment this is
```  qc::QuantumComputation qc(N);
  for (std::size_t t = 0; t < T; ++t) {
    extractSingleQubitGatesFromModel(t, model, qc, nSingleQubitGates);
    extractTwoQubitGatesFromModel(t, model, qc, nTwoQubitGates);
  }```
  
  The variables are extracted for all `t` up to `T`. But the variables only make sense for even or odd `t,` respectively.

**Review by pehamTom — COMMENTED** [review]
This already looks really good. I have given you some pointers in my review that should help identify problematic sections of the code.

In general, I would like to change the name from `STQGatesEncoder` to simply `TwoQubitEncoder`. In contrast to the `MultiGateEncoder` this should give an idea that this class is used for solving problems wrt. two-qubit gates.

File: include/cliffordsynthesis/Configuration.hpp [hunk 8]
```diff
@@ -38,6 +38,9 @@ struct Configuration {
   /// Settings for depth-optimal synthesis
   bool minimizeGatesAfterDepthOptimization = false;
 
+  /// Settings for STDepth-optimal synthesis
```
pehamTom, line 41: Comment inconsistent with variable naming

include/cliffordsynthesis/Results.hpp
```diff
@@ -40,6 +40,7 @@ class Results {
     return singleQubitGates;
   }
   [[nodiscard]] std::size_t       getDepth() const { return depth; }
+  [[nodiscard]] std::size_t       getTQDepth() const { return tQDepth; }
```
pehamTom, line 43: please rename

include/cliffordsynthesis/Results.hpp
```diff
@@ -52,6 +53,7 @@
   void setSingleQubitGates(const std::size_t g) { singleQubitGates = g; }
   void setTwoQubitGates(const std::size_t g) { twoQubitGates = g; }
   void setDepth(const std::size_t d) { depth = d; }
+  void setTQDepth(const std::size_t d) { tQDepth = d; }
```
pehamTom, line 56: please rename

include/cliffordsynthesis/Results.hpp
```diff
@@ -80,6 +82,7 @@
     resultJSON["single_qubit_gates"] = singleQubitGates;
     resultJSON["two_qubit_gates"]    = twoQubitGates;
     resultJSON["depth"]              = depth;
+    resultJSON["tQDepth"]            = tQDepth;
```
pehamTom, line 85: please rename

include/cliffordsynthesis/Results.hpp
```diff
@@ -96,8 +99,10 @@
   std::size_t       singleQubitGates = std::numeric_limits<std::size_t>::max();
   std::size_t       twoQubitGates    = std::numeric_limits<std::size_t>::max();
   std::size_t       depth            = std::numeric_limits<std::size_t>::max();
-  double            runtime          = 0.0;
-  std::size_t       solverCalls      = 0U;
+  std::size_t       tQDepth          = std::numeric_limits<std::size_t>::max();
```
pehamTom, line 102: please rename

include/cliffordsynthesis/TargetMetric.hpp
```diff
@@ -9,7 +9,7 @@
 #include <string>
 
 namespace cs {
-enum class TargetMetric { Gates, TwoQubitGates, Depth };
+enum class TargetMetric { Gates, TwoQubitGates, Depth, TQDepth };
```
pehamTom, line 12: please rename

include/cliffordsynthesis/TargetMetric.hpp
```diff
@@ -19,6 +19,8 @@
     return "two_qubit_gates";
   case TargetMetric::Depth:
     return "depth";
+  case TargetMetric::TQDepth:
+    return "tQDepth";
```
pehamTom, line 23: please rename

include/cliffordsynthesis/TargetMetric.hpp
```diff
@@ -34,6 +36,9 @@
   if (target == "depth") {
     return TargetMetric::Depth;
   }
+  if (target == "tQDepth") {
+    return TargetMetric::TQDepth;
+  }
```
pehamTom, line 41: please rename

include/cliffordsynthesis/encoding/GateEncoder.hpp
```diff
@@ -30,18 +30,24 @@
     logicbase::LogicMatrix3D gS{};
     // variables for the two-qubit gates
     logicbase::LogicMatrix3D gC{};
+    // variables for the pauli-qubit gates
+    logicbase::LogicMatrix gP{};
```
pehamTom, line 34: These variables are only used by the TwoQubitGateEncoder, please move the method there.

include/cliffordsynthesis/encoding/GateEncoder.hpp
```diff
@@ -30,18 +30,24 @@
...
 
     void
          collectSingleQubitGateVariables(std::size_t pos, std::size_t qubit,
                                          logicbase::LogicVector& variables) const;
     void collectTwoQubitGateVariables(std::size_t pos, std::size_t qubit,
                                       bool                    target,
                                       logicbase::LogicVector& variables) const;
+    void
+    collectPauliQubitGateVariables(std::size_t             qubit,
```
pehamTom, line 43: This method is only used by the TwoQubitGateEncoder, please move the method there.

include/cliffordsynthesis/encoding/GateEncoder.hpp
```diff
@@ -30,18 +30,24 @@
...
+                                   logicbase::LogicVector& variables) const;
   };
 
   // variable creation
-  void createSingleQubitGateVariables();
-  void createTwoQubitGateVariables();
+  virtual void createSingleQubitGateVariables(std::size_t tqEncoding);
+  virtual void createTwoQubitGateVariables(std::size_t tqEncoding);
```
pehamTom, line 49: Same as above, please move.

include/cliffordsynthesis/encoding/GateEncoder.hpp
```diff
@@ -52,13 +58,20 @@
...
-  static constexpr std::array<qc::OpType, 7> SINGLE_QUBIT_GATES = {
-      qc::OpType::None, qc::OpType::X, qc::OpType::Y,   qc::OpType::Z,
-      qc::OpType::H,    qc::OpType::S, qc::OpType::Sdag};
+  // TODO: set back
+  static constexpr std::array<qc::OpType, 4> SINGLE_QUBIT_GATES = {
+      qc::OpType::None, qc::OpType::H, qc::OpType::S, qc::OpType::SX};
+
+  // static constexpr std::array<qc::OpType, 7> SINGLE_QUBIT_GATES = {
+  // qc::OpType::None, qc::OpType::X, qc::OpType::Y,   qc::OpType::Z,
+  // qc::OpType::H,    qc::OpType::S, qc::OpType::Sdag};
```
pehamTom, line 71: As discussed, you can leave this like this at the moment.

include/cliffordsynthesis/encoding/GateEncoder.hpp
```diff
@@ -52,13 +58,20 @@
...
+
```
pehamTom, line 72: I think PAULI_GATES is a better name here.

include/cliffordsynthesis/encoding/SATEncoder.hpp
```diff
@@ -50,6 +50,12 @@ class SATEncoder {
     // whether to use symmetry breaking
     bool useSymmetryBreaking = false;
 
+    // whether to use SQG-TQG encoding
```
pehamTom, line 53: please rename

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -159,6 +168,9 @@ CliffordSynthesizer::determineUpperBound(EncoderConfig config) {
     if (!results.sat()) {
       lowerBound = upperBound + 1U;
       upperBound *= 2U;
+      std::cout << "No solution found for " << config.timestepLimit
+                << " timestep(s). Doubling timestep limit to " << upperBound
+                << std::endl;
```
pehamTom, line 173: Leftover Debug print. Please remove.

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -170,6 +182,8 @@
     upperBound = std::min(upperBound, results.getGates());
   } else if (config.targetMetric == TargetMetric::Depth) {
     upperBound = std::min(upperBound, results.getDepth());
+  } else if (config.targetMetric == TargetMetric::TQDepth) {
+    upperBound = std::max(upperBound, results.getTQDepth());
```
pehamTom, line 186: As we have discussed, this can't be right. The optimal synthesis result cannot have a higher depth then either of these values.

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -237,6 +251,29 @@
   }
 }
 
+void CliffordSynthesizer::twoQubitDepthOptimalSynthesis(
+    CliffordSynthesizer::EncoderConfig config, const std::size_t lower,
+    const std::size_t upper) {
+  // tQDepth-optimal synthesis is achieved by determining a timestep limit T
+  // such that there exists a solution with tQDepth T, but no solution with
+  // tQDepth T-1. This procedure uses an encoding (SQG - TQG) where in single
+  // and two qubit gates are allowed just one gate per timestep. This procedure
```
pehamTom, line 260: This is a tiny bit misleading. The number of gates is not restricted per timestep

src/cliffordsynthesis/CliffordSynthesizer.cpp
```diff
@@ -237,6 +251,29 @@
...
+  // is guaranteed to produce a depth-optimal circuit. However, the number of
+  // gates in the resulting circuit is not necessarily minimal, i.e., there may
+  // be a solution with fewer gates and the same depth. To this end, an
+  // optimization pass is provided that additionally minimizes the number of
```
pehamTom, line 264: This optimization pass is not currently present

src/cliffordsynthesis/encoding/GateEncoder.cpp
```diff
@@ -6,16 +6,17 @@
...
 #include "Encodings/Encodings.hpp"
+#include "operations/OpType.hpp"
 #include "utils/logging.hpp"
 
 namespace cs::encoding {
 
 using namespace logicbase;
 
-void GateEncoder::createSingleQubitGateVariables() {
+void GateEncoder::createSingleQubitGateVariables(const std::size_t tqEncoding) {
```
pehamTom, line 16: The GateEncoder class should not have to care whether a two-qubit encoding is used. That breaks the separation of concerns that inheritance tries to provide. Overriding the method in the TwoQubitEncoder class is the better solution.
Also, tqEncoding should be a `bool`

src/cliffordsynthesis/encoding/GateEncoder.cpp
```diff
@@ -31,10 +32,10 @@
   }
 }
 
-void GateEncoder::createTwoQubitGateVariables() {
+void GateEncoder::createTwoQubitGateVariables(const std::size_t tqEncoding) {
```
pehamTom, line 35: same as above

src/cliffordsynthesis/encoding/GateEncoder.cpp
```diff
@@ -51,6 +52,21 @@
...
+    g.reserve(N);
+    for (std::size_t q = 0U; q < N; ++q) {
+      const std::string gName = "g_" + std::to_string(T) + "_" +
+                                toString(gate) + "_" + std::to_string(q);
+      TRACE() << "Creating variable " << gName;
+      g.emplace_back(lb->makeVariable(gName));
+    }
+  }
+}
+
```
pehamTom, line 69: As mentioned in the header, move this to TwoQubitEncoder class.

src/cliffordsynthesis/encoding/GateEncoder.cpp
```diff
@@ -77,6 +93,13 @@
   }
 }
 
+void GateEncoder::Variables::collectPauliQubitGateVariables(
+    const std::size_t qubit, LogicVector& variables) const {
+  for (const auto& gate : gP) {
+    variables.emplace_back(gate[qubit]);
+  }
+}
+
```
pehamTom, line 102: Same as above

src/cliffordsynthesis/encoding/SATEncoder.cpp
```diff
@@ -55,6 +55,10 @@ void SATEncoder::createFormulation() {
                             ? 2U * N
                             : N;
 
+  if (config.useTwoQubitEncoding) {
+    T = T * 2U + 1U;
```
pehamTom, line 59: As discussed, be careful with overflows here

src/cliffordsynthesis/encoding/SATEncoder.cpp
```diff
@@ -63,15 +67,25 @@
...
   }
-  gateEncoder->createSingleQubitGateVariables();
-  gateEncoder->createTwoQubitGateVariables();
+
   gateEncoder->encodeGates();
 
-  if (config.useSymmetryBreaking) {
+  if (config.useSymmetryBreaking && !config.useTwoQubitEncoding) {
     gateEncoder->encodeSymmetryBreakingConstraints();
   }
```
pehamTom, line 90: Will have to be changed once methods are adjusted.
Also, you don't have to check for `useTwoQubitEncoding` if you make the `useSymmetryBreaking` method just not do anything. Then we don't have this strange special case here and don't have to adapt anything if somebody decides to implement symmetry breakers in the future.

src/cliffordsynthesis/encoding/TwoQubitEncoder.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+#include "LogicTerm/LogicTerm.hpp"
+#include "operations/OpType.hpp"
+#include "utils/logging.hpp"
+
+namespace cs::encoding {
+
+using namespace logicbase;
+
+void encoding::TwoQubitEncoder::collectTwoQubitGateVariables(
```
pehamTom, line 15: This method can be simplified by calling the method from the base class which already adds all variables except the `twoQubitGates[qubit][qubit]` one

src/cliffordsynthesis/encoding/TwoQubitEncoder.cpp
```diff
@@ -0,0 +1,251 @@
...
+    const std::size_t pos, const std::size_t qubit, const bool target,
+    LogicVector& variables) const {
+  const auto& twoQubitGates = vars.gC[pos];
+  const auto  n             = twoQubitGates.size();
+  for (std::size_t q = 0; q < n; ++q) {
+    if (q == qubit) {
+      if (!target) {
+        variables.emplace_back(twoQubitGates[qubit][qubit]);
```
pehamTom, line 23: A comment as to why this variable is treated special might be helpful

src/cliffordsynthesis/encoding/TwoQubitEncoder.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+void encoding::TwoQubitEncoder::assertConsistency() const {
+  DEBUG() << "Asserting gate consistency";
+  for (std::size_t t = 0U; t <= T; ++t) {
+    // asserting only a single gate is applied on each qubit.
+    for (std::size_t q = 0U; q < N; ++q) {
+      LogicVector singleQubitGateVariables{};
+      LogicVector twoQubitGateVariables{};
+      LogicVector pauliGateVariables{};
+      if (t == T) {
```
pehamTom, line 43: This does not have to be done in the loop since it is indepented of `t`

src/cliffordsynthesis/encoding/TwoQubitEncoder.cpp
```diff
@@ -0,0 +1,251 @@
...
+        vars.collectPauliQubitGateVariables(q, pauliGateVariables);
+        assertExactlyOne(pauliGateVariables);
+        IF_PLOG(plog::verbose) {
+          TRACE() << "Pauli Qubit Gate variables at time " << t << " and qubit "
+                  << q;
+          for (const auto& var : pauliGateVariables) {
+            TRACE() << var.getName();
+          }
+        }
+      } else if (t % 2 == 0) {
```
pehamTom, line 53: The case distinction here is redundant since `t/2` will give the same result in both cases. Simply adjust the loop to move in steps of `2` (`t+=2`) or set the bound to `T/2`.

src/cliffordsynthesis/encoding/TwoQubitEncoder.cpp
```diff
@@ -0,0 +1,251 @@
...
+          for (const auto& var : twoQubitGateVariables) {
+            TRACE() << var.getName();
+          }
+        }
+      }
+    }
+  }
+}
+
+void encoding::TwoQubitEncoder::assertGateConstraints() {
```
pehamTom, line 82: After playing around with the code a bit I am now 100% convinced that a last single-qubit gate layer is need after the last two-qubit gate layer. Otherwise depth-optimality cannot be guaranteed.

src/cliffordsynthesis/encoding/TwoQubitEncoder.cpp
```diff
@@ -0,0 +1,251 @@
...
+  DEBUG() << "Asserting gate constraints";
+  xorHelpers = logicbase::LogicMatrix{T + 1};
+  for (std::size_t t = 0U; t <= T; ++t) {
+    TRACE() << "Asserting gate constraints at time " << t;
+    splitXorR(tvars->r[t], t);
+    if (t == T) {
+      assertPauliGateConstraints(t);
```
pehamTom, line 89: Again, this doesn't need to be in the loop.

src/cliffordsynthesis/encoding/TwoQubitEncoder.cpp
```diff
@@ -0,0 +1,251 @@
...
+    for (const auto gate : PAULI_QUBIT_GATES) {
+      const auto& change =
+          LogicTerm::ite(vars.gP[gateToIndex(gate)][q],
+                         tvars->singleQubitRChange(pos, q, gate),
+                         LogicTerm(0, static_cast<std::int16_t>(S)));
+
+      splitXorR(change, pos);
+    }
+    lb->assertFormula(tvars->x[pos][q] == tvars->x[pos + 1][q]);
+    lb->assertFormula(tvars->z[pos][q] == tvars->z[pos + 1][q]);
```
pehamTom, line 145: Not necessary, but since only the phase is adjusted here, one could save on Tableau Variables.

src/cliffordsynthesis/encoding/TwoQubitEncoder.cpp
```diff
@@ -0,0 +1,251 @@
...
+            LogicTerm::implies(twoQubitGates[ctrl][trgt], changes));
+        DEBUG() << "Asserting CNOT on " << ctrl << " and " << trgt;
+      }
+    }
+  }
+}
+
+LogicTerm
+encoding::TwoQubitEncoder::createIdentityConstraintOnTQG(std::size_t pos,
+                                                         std::size_t ctrl) {
```
pehamTom, line 187: This does not have to be a method. It is simple enough to be inlined and is only called once.

src/cliffordsynthesis/encoding/TwoQubitEncoder.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+  qc::QuantumComputation qc(N);
+  for (std::size_t t = 0; t < T / 2; ++t) {
+    extractSingleQubitGatesFromModel(t, model, qc, nSingleQubitGates);
+    extractTwoQubitGatesFromModel(t, model, qc, nTwoQubitGates);
+  }
+
+  res.setSingleQubitGates(nSingleQubitGates);
+  res.setTwoQubitGates(nTwoQubitGates);
+  res.setTQDepth(qc.getTQDepth());
```
pehamTom, line 226: This should be for the other encodings as well so you can compare against them. Add this computation to `GateEncoder::extractCircuitFromModel`

pehamTom (commented):
Thank you for your work. I have reviewed your code and pointed out changes I'd like to see.
As stated in the comments you will most certainly need another single-qubit gate layer after the last two-qubit gate layer.
The naming TQDepth is not very self-explainatory. Please rename the respective variables to twoQubitDepth...

burgholzer:
Closing this due to inactivity. It's been two years. @pehamTom please feel free to reopen any time if you plan to keep working on this.

PR #397, opened 2023-11-14
Delayed Phase Correction

### LINKED ISSUE #326: ✨ Delay Paulis until the end of the circuit.
### What's the problem this feature will solve?

Pauli gates can always be pushed to the end of the circuit when synthesizing Clifford circuits. At the moment Pauli gates are explicitely considered at every depth in the encoding which unnecessarily increases the size of the formulation and the search space.

### Describe the solution you'd like

Since the Pauli gates only ever change the Phase of the stabilizer tableau, all constraints on the phase and the phase variables themselves can be dropped in the encoding. After circuit synthesis, the phases can be corrected by simulating the circuit and flipping phases with the appropriate Pauli gate.

pehamTom:
## Description

This PR introduces a feature where stabilizer phases are ignored during Clifford synthesis. If that happens, the phases are corrected during a post-processing step. 

Fixes #326

## Checklist:

- [ ] The pull request only contains commits that are related to it.
- [ ] I have added appropriate tests and documentation.


burgholzer:
Hey @pehamTom 👋🏼 

Haven't looked through this at all yet. Just a brief question: Is this something that's always guaranteed to be faster than the original version? If so, could we just fully replace the old version and avoid adding all those if statements for deciding between both options? Just thinking about maintainability here.

pehamTom:
Ignoring the phase during synthesis does not guarantee depth-optimality. One could always back-propagate Paulis and fuse them with any single-qubit gates as a post-processing step. But I think it makes sense to respect the chosen gateset and leave the rest up to the user. 

I wanted to compress the settings we have anyway. The maxsat method should be removed entirely anyway as it is always worse. But that is something for another PR.

burgholzer:
> Ignoring the phase during synthesis does not guarantee depth-optimality. One could always back-propagate Paulis and fuse them with any single-qubit gates as a post-processing step. But I think it makes sense to respect the chosen gateset and leave the rest up to the user.
> 
> I wanted to compress the settings we have anyway. The maxsat method should be removed entirely for example as it is always worse. But that is something for another PR.

Sorry for responding so late here. For some reason, I missed the notification for the message.

I see! In that case, the separate option definitely makes sense.
Compressing the settings is definitely much appreciated (in a separate PR, of course).

Is this PR ready for review then?

pehamTom:
Now it is ready for review.

src/cliffordsynthesis/encoding/SATEncoder.cpp
```diff
@@ -50,22 +55,30 @@ void SATEncoder::createFormulation() {
...
-  tableauEncoder = std::make_shared<TableauEncoder>(N, s, T, lb);
+  tableauEncoder =
+      std::make_shared<TableauEncoder>(N, S, T, lb, config.ignoreRChanges);
   tableauEncoder->createTableauVariables();
   tableauEncoder->assertTableau(*config.initialTableau, 0U);
   tableauEncoder->assertTableau(*config.targetTableau, T);
 
+  if (config.ignoreRChanges) {
+    config.gateSet.removePaulis();
+    config.gateSet.emplace_back(qc::OpType::None);
```
burgholzer, line 71: Multiple times throughout this PR I had the feeling that `None` is not consistently handled. If you have to add it back anyways, you could just not remove it in the first place? There is a lot of removing and adding `None` in the gate set handling that I believe could be consolidated and mostly eliminated.

src/cliffordsynthesis/encoding/SATEncoder.cpp
```diff
@@ -114,7 +127,38 @@ Result SATEncoder::solve() const {
 void SATEncoder::extractResultsFromModel(Results& res) const {
   auto* const model = lb->getModel();
   tableauEncoder->extractTableauFromModel(res, T, *model);
-  gateEncoder->extractCircuitFromModel(res, *model);
+  auto qc = gateEncoder->extractCircuitFromModel(res, *model);
```
burgholzer, line 130: This might create an unnecessary copy of the quantum circuit in case `ignoreRChanges=false`. It could be that RVO elides that copy, but I wouldn't be 100% sure.
Would it be cleaner to refactor the `Results` class so that it does not just store the OpenQASM string of the circuit and the string representation of the tableau, but the actual circuit and tableau? It would then produce the QASM and Tableau string only when requested.

src/cliffordsynthesis/encoding/SATEncoder.cpp
```diff
@@ -114,7 +127,38 @@ Result SATEncoder::solve() const {
...
+  if (config.ignoreRChanges) {
+    const auto start = std::chrono::high_resolution_clock::now();
+    auto       tab   = *config.targetTableau;
+    auto       qcTab = Tableau(qc, 0, std::numeric_limits<std::size_t>::max(),
+                               config.targetTableau->hasDestabilizers());
```
burgholzer, line 135: If you were to implement the change proposed above, I would guess that this can be skipped completely as the resulting tableau from the circuit is available in the results.

src/python/bindings.cpp
```diff
@@ -4,17 +4,36 @@
...
+    if (l.size() > 0) {
+      for (auto& gate : l) {
+        if (py::isinstance<py::str>(gate)) {
+          gs.emplace_back(qc::opTypeFromString(gate.cast<std::string>()));
+        }
+      }
+    }
+  }
+  return gs;
+}
```
burgholzer, line 35: if this only works with lists, then the function signature should reflect that. It would be nice to explore whether pybind even allows you to request a list of strings. This the function as it is could be called with quite some garbage.
```suggestion
cs::GateSet loadGateSet(const py::list& gateSet) {
  cs::GateSet gs{};
  if (gateSet.size() > 0) {
    for (auto& gate : gateSet) {
      if (py::isinstance<py::str>(gate)) {
        gs.emplace_back(qc::opTypeFromString(gate.cast<std::string>()));
      }
    }
  }
  return gs;
}
```

src/python/bindings.cpp
```diff
@@ -504,6 +523,12 @@ PYBIND11_MODULE(pyqmap, m) {
           "n_threads_heuristic", &cs::Configuration::nThreadsHeuristic,
           "Maximum number of threads used for the heuristic optimizer. "
           "Defaults to the number of available threads on the system.")
+      .def_readwrite("gate_set", &cs::Configuration::gateSet,
+                     "Gate Set to be used for the Synthesis. "
+                     "Defaults to {H, S, Sdg, X, Y, Z, CX}.")
```
burgholzer, line 528: This comment is not really true. `CX` newer really plays into whatever is being set here and there is no way to specify other two-qubit gates. Guess this is mostly a consequence of the inappropriate naming.

src/python/bindings.cpp
```diff
@@ -504,6 +523,12 @@ PYBIND11_MODULE(pyqmap, m) {
...
+      .def_readwrite("delay_paulis", &cs::Configuration::delayPaulis,
+                     "Delay Pauli gates to the end of the circuit. "
+                     "Defaults to `false`.")
```
burgholzer, line 531: I believe it would be good to add a statement here that comments on whether this preserves optimality.

src/python/bindings.cpp
burgholzer, line 1: All of the new options are missing from the Python stubs for QMAP. Please also add any additions to the bindings there.

test/cliffordsynthesis/test_synthesis.cpp
burgholzer, line 1: I am surprised by the lack of further tests asserting the proper behaviour given the size of the changes in this PR.
Is this really enough to ensure that this works as expected?
For example: you introduce single-qubit gate sets in this PR, but you newer test different gate-sets.
That seems very dangerous to me to introduce regressions.

src/cliffordsynthesis/encoding/PhaseCorrectionEncoder.cpp
```diff
@@ -0,0 +1,148 @@
...
+    LogicVector yChange{};
+    iChange.reserve(S);
+    xChange.reserve(S);
+    zChange.reserve(S);
+    yChange.reserve(S);
+
+    for (std::size_t row = 0U; row < S; ++row) {
+      DEBUG() << "Encoding identity constraint.";
+      iChange.emplace_back(logicbase::LogicTerm::ite(
+          paulis[q][0], LogicTerm(false), LogicTerm(false)));
```
burgholzer, line 61: Something seems off to me about that constraint no matter the value of the Paulis, this will always be false, right?
Like, this can be way simpler and outside the loop.

src/cliffordsynthesis/encoding/PhaseCorrectionEncoder.cpp
```diff
@@ -0,0 +1,148 @@
...
+      lb->assertFormula(xorHelpers[0][row] == changes[row]);
+    }
+  } else {
+    for (std::size_t row = 0U; row < S; ++row) {
+      lb->assertFormula(
+          xorHelpers.back()[row] ==
+          (changes[row] != xorHelpers[xorHelpers.size() - 2][row]));
+    }
+  }
+}
```
burgholzer, line 109: This method is already implemented elsewhere if I am not mistaken. Can we avoid the code duplication here?

src/cliffordsynthesis/encoding/PhaseCorrectionEncoder.cpp
```diff
@@ -0,0 +1,148 @@
...
+  return result;
+}
+
+std::vector<qc::OpType> PhaseCorrectionEncoder::extractResult() {
+  const auto&             model = lb->getModel();
+  std::vector<qc::OpType> result{};
+  result.reserve(N);
+  for (std::size_t q = 0U; q < N; ++q) {
+    GateSet pauliGates{};
+    pauliGates.removePaulis();
```
burgholzer, line 127: This is a truly strange sequence of calls to me that only makes me ask: WHY? 🙃 
You create an empty gate set and then you remove Paulis from that?

burgholzer (changes requested):
Many thanks for submitting this PR!

For future reference: It would have been appreciated if, instead of a large PR like this that accomplishes more than a single thing, this were a collection of smaller self-contained PRs. That would make it so much easier to reason about and would considerably speed up the adoption of changes in `main`. 
I guess it's too late for that now in this case.
That's one of the main reasons why the list of comments below is rather long. Sorry for that.

You should be able to batch the smaller code suggestions together directly on GitHub and commit them in one go. That should get rid of some comments already.

Two things that I couldn't really annotate in the sources themselves:
- Please update the PR description to reflect all the additions, optimisations, changes in this PR
- Is this a feature that should be advertised in the documentation? For example in the existing Clifford Synthesis notebook?

burgholzer:
Closing this due to inactivity. @pehamTom please feel free to reopen any time if you plan to keep working on this. I still believe this would be a nice feature.

PR #405, opened 2023-12-23
Fixing minor bug in `disjoint_2q_block` Layering and adding option `add_barriers_between_layers`

EliasLF:
## Description

This PR fixes a minor bug in `disjoint_2q_block` Layering, where it previously pushed 2q-gates erroneously to the second layer, if the first layer has 1q-gates already applied to the same qubits. 

I also took this as an opportunity to separate `disjoint_2q_block` and `disjoint_qubits` into 2 different functions for the sake of readability. It was already getting cluttered now, and with a soon-to-come suggestion for another similar layering technique (specifically designed for noise-aware mapping) the problem will probably only get worse. 
In my opinion, the ensuing code duplication is not too bad, but if you disagree, I am also happy to reintegrate both functions back into one.

Additionally, I added the mapping option `add_barriers_between_layers` intended for making layer boundaries visible after the mapping process.

## Checklist:

### DIFF
--- a/src/Mapper.cpp
+++ b/src/Mapper.cpp
@@ -35,18 +35,11 @@ Mapper::Mapper(qc::QuantumComputation quantumComputation, Architecture& arch)
 void Mapper::processDisjointQubitLayer(
     std::array<std::optional<std::size_t>, MAX_DEVICE_QUBITS>& lastLayer,
     const std::optional<std::uint16_t>& control, const std::uint16_t target,
-    qc::Operation* gate, bool collect2qBlocks) {
+    qc::Operation* gate) {
   std::size_t layer = 0;
   if (!control.has_value()) {
     if (lastLayer.at(target).has_value()) {
-      layer = *lastLayer.at(target);
-      // single qubit gates can always be added to the last 2Q block
-      if (!collect2qBlocks) {
-        layer++;
-      }
-    }
-    if (!collect2qBlocks) {
-      lastLayer.at(target) = layer;
+      layer = *lastLayer.at(target) + 1;
     }
     lastLayer.at(target) = layer;
   } else {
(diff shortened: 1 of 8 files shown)


src/Mapper.cpp
```diff
@@ -35,20 +35,49 @@ Mapper::Mapper(qc::QuantumComputation quantumComputation, Architecture& arch)
...
+}
+
+void Mapper::processDisjoint2qBlockLayer(
+    std::array<std::optional<std::size_t>, MAX_DEVICE_QUBITS>& lastLayer,
+    const std::optional<std::uint16_t>& control, const std::uint16_t target,
+    qc::Operation* gate) {
+  std::size_t layer = 0;
+  if (!control.has_value()) {
+    // single qubit gates can always be added to the last 2Q block and should
+    // affect placings of future 2Q blocks
```
burgholzer, line 77:
```suggestion
    // single qubit gates can always be added to the last 2Q block and should
    // not affect placings of future 2Q blocks
```
Could it be that a "not" is missing here?

test/test_heuristic.cpp
```diff
@@ -72,6 +72,91 @@ TEST(Functionality, NodeCostCalculation) {
...
+  // check barrier count
+  barriers = 0;
+  for (const auto& op : qcMapped) {
+    if (op->getType() == qc::Barrier) {
+      ++barriers;
+    }
+  }
+  EXPECT_EQ(barriers, result.input.layers);
+}
+
```
burgholzer, line 159: This is really just a minor suggestion so feel free to simply object, but it could be worth turning this into a set of tests with the circuit and configuration as a fixture.
Not that it's too critical, but I guess it's always better to have more, but smaller tests. Especially given that they can be run in parallel.

burgholzer (approved):
Many thanks. Looking good ✅
Stumbled over two minor things that should be easily addressed or disregarded.

src/Mapper.cpp
EliasLF: That's true. Thank you for catching that.

test/test_heuristic.cpp
EliasLF: I agree. Thanks for the suggestion!

PR #410, opened 2024-01-02
Restructuring of the heuristic system

EliasLF:
## Description

This PR strives to solve multiple problems (that are just too interconnected to keep them to separate PRs):
- 2 new heuristics `GateCountSumDistanceMinusSharedSwaps` and `GateCountMaxDistanceOrSumDistanceMinusSharedSwaps`, where the former is the sum of _all_ qubit pair distances minus an upper bound for how many swaps could potentially be saved by sharing with other moving qubits, and the latter is just the dominating heuristic over the previous `GateCountMaxDistance` and the new `GateCountSumDistanceMinusSharedSwaps`. 
Tested on a subset of MQTbench and mapping to IBM Brisbane `GateCountMaxDistanceOrSumDistanceMinusSharedSwaps` reduces the effective branching rate by about 10% compared to `GateCountMaxDistance`, while of course yielding the same results (if lookahead is disabled) as both heuristics are principally admissible. 
Unfortunately, the current default lookahead heuristic `GateCountMaxDistance` seems to not work well with this new heuristic, as the combination results in slightly higher costs. This is probably due to scaling issues (since the new heuristic is closer to the real cost and therefore larger, reducing the impact of the lookahead penalty) but might be solvable with higher lookahead factors or a new better suited lookahead heuristic.
- Introducing a more flexible system for heuristics (both for the main heuristic and the lookahead heuristic). In the new system any implementation specifics of a heuristic are isolated to a single function calculating a heuristic value from a search node, outside of those functions only a few characteristics of the heuristics are relevant for the mapper: 
  - principal admissibility (i.e. admissibility at least on the optimal solution path)
  - tightness (heuristics that are 0 in all goal nodes)
  - fidelity-awareness
- Fixing the handling of CNOT reversals in Dijkstra, fixed-cost-calculation and all the pre-existing heuristics. The current implementation resulted in both non-admissible heuristics and fixed costs that did not accurately reflect the gates added by QMAP (the cost calculation currently assumes that at most 1 reversal is added, while QMAP actually inserts reversals for each backwards CNOT). This new implementation is as robust as possible, only failing to produce admissible heuristics in edge-cases where cumulative reversal costs on one edge surpass SWAP costs resulting in a non-convex cost space (which generally does not allow for heuristics, that are both admissible and tight)
- Moving all methods from `HeuristicMapper::Node` to `HeuristicMapper` (since they grew more and more dependent on data structures from `HeuristicMapper` with recent PRs) and making them more atomic (i.e. reducing points at which nodes are in an inconsistent state)
- Added support for semi-directional architectures (i.e. architectures with both bidirectional and unidirectional edges)

List of minor changes and bug fixes:
- Cleaning up `HeuristicMapper::Node`:
  - Removing `nswaps` because of redundancy with `swaps.size()`
  - Making `swaps` a one-dimensional vector (I assume the inner dimension was originally intended for the possibility of adding multiple swaps per node. However, this is currently not implemented, so all inner vectors are just of length 1)
  - renaming `done` to `validMapping`, which better describes the property now that non-tight heuristics have been introduced to QMAP
- Fixed tracking of `HeuristicMapper::Node::validMappedTwoQubitGates` and activated it also for non-fidelity-aware heuristics enabling a more efficient check if a node has a valid mapping.
- Reduced the register sizes in the example circuits to the actual number of qubits used in each circuit, to allow for easier checks in the tests for the minimum required architecture size
- Made the ordering of validly mapped search nodes consistent (as defined by `operator>`), where previously all validly mapped nodes with the same total cost were considered equal (resulting in an arbitrary ordering in the priority queue and thereby potentially different optimal solutions for different principally admissible heuristics)
- fixing a bug in the handling of teleportation qubits in `HeuristicMapper::createInitialMapping`, that causes this function to go into an endless loop in some edge cases (instead of randomly iterating over the whole coupling map, it only iterates over a part of it due to the upper limit for the RNG being too low, which causes an endless loop once that subgraph is fully mapped but free teleportation qubits remain)
- removing redundant `Mapper::fidelities`
- simplifying `Dijkstra::buildEdgeSkipTable` by internally calling `Dijkstra::buildTable` for the 0th dimension
- moving checks for invalid settings from `HeuristicMapper::map` to new method `HeuristicMapper::checkParameters`

## Checklist:


include/heuristic/HeuristicMapper.hpp
```diff
@@ -49,125 +49,101 @@ class HeuristicMapper : public Mapper {
...
+         const std::size_t initSharedSwaps                                = 0)
+        : costFixed(initCostFixed), costFixedReversals(initCostFixedReversals),
+          sharedSwaps(initSharedSwaps), depth(searchDepth), parent(parentId),
+          id(nodeId) {
       std::copy(q.begin(), q.end(), qubits.begin());
       std::copy(loc.begin(), loc.end(), locations.begin());
       std::copy(sw.begin(), sw.end(), std::back_inserter(swaps));
+      std::copy(valid2QGates.begin(), valid2QGates.end(),
+                std::inserter(validMappedTwoQubitGates,
+                              validMappedTwoQubitGates.begin()));
```
burgholzer, line 106: I guess this is one of those points where I have lost the original intention behind the code. Why are these 4 parameters here copied explicitly with `std::copy` and not just using the regular copy constructor?

include/heuristic/HeuristicMapper.hpp
```diff
@@ -275,28 +240,174 @@ class HeuristicMapper : public Mapper {
...
+   * @brief calculates the heuristic using
+   * `Heuristic::FidelityBestLocation`
+   *
+   * @param layer index of current circuit layer
+   * @param node search node for which to calculate the heuristic cost
+   *
+   * @return heuristic cost
+   */
+  double heuristicFidelityBestLocation(std::size_t layer, Node& node);
+
```
burgholzer, line 377: Another side comment/observation: I could be worth exploring whether we could turn the Heuristics from an enum to a proper class (hierarchy) so that methods like these could be bundled into a class and would not be so free-floating and loosely tied together. Same goes for the Lookahead strategies down below.
Definitely nothing for this PR though,

include/heuristic/HeuristicMapper.hpp
```diff
@@ -341,16 +452,23 @@ inline bool operator<(const HeuristicMapper::Node& x,
...
   }
 
-  if (x.done) {
+  if (x.validMapping && !y.validMapping) {
     return false;
   }
-  if (y.done) {
+  if (y.validMapping && !x.validMapping) {
     return true;
   }
```
burgholzer, line 474: Technically this could be simplified to something like 
```c++
if (x.validMapping != y.validMapping) {
  return y.validMapping;
}
```

src/Architecture.cpp
```diff
@@ -185,21 +185,34 @@ Architecture::Architecture(const std::uint16_t nQ, const CouplingMap& cm,
...
+  Dijkstra::buildSingleEdgeSkipTable(simpleDistanceTable, couplingMap, 0.,
+                                     distanceTable);
+  if (bidirectional()) {
+    distanceTableReversals = distanceTable;
+  } else {
+    Dijkstra::buildSingleEdgeSkipTable(simpleDistanceTable, couplingMap,
+                                       COST_DIRECTION_REVERSE,
+                                       distanceTableReversals);
+  }
 }
```
burgholzer, line 216: This has gotten quite a bit more complicated then before. Wouldn't it make sense to move this to where the Dijkstra routines are located so that it only becomes a single call here?
Also, I am still a bit puzzled by the need for the additional table, but I guess that will resolve after you had time to comment.

src/configuration/Configuration.cpp
```diff
@@ -19,18 +19,18 @@ nlohmann::json Configuration::json() const {
...
-    heuristic["admissible_heuristic"] = admissibleHeuristic;
-    heuristic["consider_fidelity"]    = considerFidelity;
-    if (lookahead) {
-      auto& lookaheadSettings           = heuristic["lookahead"];
+    auto& heuristicJson             = config["settings"];
+    heuristicJson["heuristic"]      = ::toString(heuristic);
+    heuristicJson["initial_layout"] = ::toString(initialLayout);
+    if (lookaheadHeuristic != LookaheadHeuristic::None) {
+      auto& lookaheadSettings           = heuristicJson["lookahead"];
+      lookaheadSettings["heuristic"]    = ::toString(lookaheadHeuristic);
```
burgholzer, line 27: This could be the perfect place to add some more metadata from the functions you marked as `maybe_unused`. It could be helpful to know whether the selected heuristic is admissible, tight, etc.

src/heuristic/HeuristicMapper.cpp
```diff
@@ -7,29 +7,36 @@
...
-    throw QMAPException("Lookahead is not yet supported for heuristic mapper "
-                        "using fidelity-aware mapping!");
+  if (fidelityAwareHeur && !isFidelityAware(config.lookaheadHeuristic) &&
+      config.lookaheadHeuristic != LookaheadHeuristic::None) {
+    throw QMAPException("Fidelity-aware heuristics may only be used with "
+                        "fidelity-aware lookahead heuristics (or no "
+                        "lookahead)!");
   }
-  if (config.considerFidelity && config.teleportationQubits > 0) {
+  if (fidelityAwareHeur && config.teleportationQubits > 0) {
```
burgholzer, line 39: At some point, we should probably refactor all these checks into a separate function like `checkParameters` (or similar) so that these additional conditions do not further bloat the `map` function.

src/heuristic/HeuristicMapper.cpp
```diff
@@ -1025,198 +1076,568 @@ void HeuristicMapper::Node::applyTeleportation(const Edge&   swap,
...
+    // infer maximum number of swaps in this distance
+    if (architecture->unidirectional()) {
+      nSwaps.emplace_back(
+          static_cast<std::size_t>(swapCost / COST_UNIDIRECTIONAL_SWAP));
+    } else {
+      nSwaps.emplace_back(
+          static_cast<std::size_t>(swapCost / COST_BIDIRECTIONAL_SWAP));
+    }
+  }
+
```
burgholzer, line 1349: Just a few lines below you use `architecture->bidirectional` for the if condition. This seems a little inconsistent and is probably one of the reasons why I originally brought it up in the comment for the changes in the `Architecture` class.

test/test_heuristic.cpp
```diff
@@ -12,64 +12,935 @@
...
+
+  architecture->loadCouplingMap(5, {{0, 1}, {1, 2}, {3, 1}, {4, 3}});
+  qc = qc::QuantumComputation{5};
+  // layer 0 distances:
+  // 0-1: 2 swaps & 2 reversals
+  // 2-3: 0 swaps & 1 reversals
+  qc.cx(qc::Control{0}, 1);
+  qc.cx(qc::Control{0}, 1);
+  qc.cx(qc::Control{0}, 1);
+  qc.cx(qc::Control{0}, 1);
```
burgholzer, line 215: Just for reference. I think you should be able to get away with `qc.cx(0,1)` (or, potentially `qc.cx(0U, 1U)`)

burgholzer (changes requested):
Many thanks for another great contribution. What a nice way to start the year!
I really like the direction this PR is taking. Makes the code much more organized.
Most of the comments below are really just comments. Feel free to read them, think about them, and discard them if you don't agree with them.
I think the only point where I might not be that happy is the `Architecture` class as some redundancy seems to be introduced by the PR. We should be able to clarify this quickly though!

EliasLF:
Alright, first of all thank you very much for the quick review! 

Some of the design choices in this PR were made with future plans/issues in mind, which are (or were previously) not mentioned in the PR description above. Before going into detail in your code comments, maybe let's first discuss them in general here:
- semi-directional architectures (i.e. architectures with both bi- and unidirectional edges)
- user-defined gate costs (for the non-fidelity-aware case)
- arbitrary 2q gates

If I'm not mistaken, we already talked about all of these in person but it's probably a good idea to also get this into writing here on Github:

The order above is (most likely) also the order of their relevancy (from low to high) and unfortunately also the order of the incurred cost/complexity (from low to high). 

To my knowledge currently there exist no semi-directional architectures in practice. However, our heuristic mapper is implemented abstractly enough, that allowing for such architectures comes almost for free (2 extra jump instructions per search node (for purely bi-/unidirectional architectures), a boolean field `Architecture::isUnidirectional`, a getter method for that field and 1 small else-case in the loop of `Architecture::createDistanceTable`).
Originally, the plan was to tackle this issue in a future PR, but thanks to your comments I just realized that there were only 2 lines remaining (in the heuristic mapper) that assumed pure directionality. I therefore just pushed these remaining changes (+ a few optimizations) and added this "feature" to the PR description.
If you disagree with the decision to allow semi-directional architectures, it would be quite easy to revert the changes. In that case, however, I think we should at least specifically check and disallow loading such coupling maps (which currently is not the case, they are just not correctly handled during mapping)

User-defined gate costs and arbitrary 2q gates are probably much more relevant, but also more complex, which I only realized recently. As mentioned in the PR description, even the current implementation strictly speaking wrongly assumes a convex cost space due to cumulative reversal costs on logical edges possibly surpassing swap costs (in the non-fidelity-aware case on non-bidirectional architectures). E.g. if there are 9 congruent 2q-gates in a layer, which are already validly mapped to a back-edge incurring 9*4=36 in reversal costs, swapping them to a neighboring forward-edge with 1 swap only costs 34, i.e. the first goal node on the search path is not the optimal goal node. 
This is currently not a huge problem, since it's rare to find 9 congruent CNOTs in 1 layer (and even only possible with `Disjoint2qBlock` layering) and could easily be optimized down to at most 3 CNOTs in the pre-optimization stage. However, with more evened out costs between H and CNOT, or different reversal operations for other 2q gates, this non-convexity might become more significant.
The new distance/dijkstra system (not only fixes the currently flawed, non-admissible system, but also) gives flexibility for future heuristics tackling this non-convexity problem by dropping the tightness constraint. In contrast to the previous custom dijkstra solution, edge-skipping dijkstra will return the correct distance for any combination of swap cost and reversal cost; and 1 execution with 1 skip (as it is currently used in the non-fidelity-aware case) it only increases the runtime complexity from O(E*log V) to O(E*log V + V²*E), which is not too bad for something that runs only once per mapping.

include/Architecture.hpp
```diff
@@ -476,12 +489,14 @@ class Architecture {
...
-  CouplingMap                                        couplingMap           = {};
-  CouplingMap                                        currentTeleportations = {};
-  bool                                               isBidirectional = true;
-  Matrix                                             distanceTable   = {};
+  std::string   name;
+  std::uint16_t nqubits                = 0;
+  CouplingMap   couplingMap            = {};
+  CouplingMap   currentTeleportations  = {};
+  bool          isBidirectional        = true;
+  bool          isUnidirectional       = true;
```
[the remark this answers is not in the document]
EliasLF: `isUnidirectional` is needed to handle semi-directional architectures without much overhead for bi- and unidirectional architectures. (see my comment on the main thread for more info)
To the question of how they can be both true: This is only the case for an empty coupling map. I use the following definition (which is the most convenient for the heuristic mapper):
An architecture is bidirectional if there are no unidirectional edges, and it is unidirectional if there are no bidirectional edges.

Therefore only the empty coupling map (which is, however, how `Architecture` is initialized per default) has `isBidirectional && isUnidirectional`. For semidirectional architectures both are `false` and for bidirectional and unidirectional architectures only one of these is set to true.

A comment explaining this, however, might probably be a good idea.

include/Architecture.hpp
```diff
@@ -406,9 +416,12 @@ class Architecture {
...
   }
 
-  [[nodiscard]] double distance(std::uint16_t control,
-                                std::uint16_t target) const {
+  [[nodiscard]] double distance(std::uint16_t control, std::uint16_t target,
+                                bool includeReversalCost = true) const {
     if (currentTeleportations.empty()) {
+      if (includeReversalCost) {
+        return distanceTableReversals.at(control).at(target);
+      }
```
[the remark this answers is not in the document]
EliasLF: I thought the naming was clear enough here (should I add doc strings for those tables?). `distanceTable` contains just the swap distance (i.e. the cheapest path distance from 1 qubit to another qubit skipping 1 edge, on which the respective gate is executed) and `distanceTableReversals` contains the swap distance including the minimum cost for reversals (currently at most 1 reversal is considered). The former is needed in the new `GateCountSumDistanceMinusSharedSwaps` heuristic, which has to handle gate reversals separately to ensure principal admissibility.

I think, given only purely bi-/unidirectional architectures, the current gate costs and as long as 1 reversal remains the upper limit we consider, we could also just use only `distanceTable` and add COST_DIRECTION_REVERSE if needed, since the cheapest path should not change in that case, but I would argue that the added flexibility is definitely worth the second table (probably not for the unrealistic use-case of semi-directional architectures, but definitely for variable gate costs and arbitrary 2q gates)

include/configuration/Heuristic.hpp
```diff
@@ -0,0 +1,158 @@
...
+  FidelityBestLocation
+};
+
+/**
+ * A heuristic is admissible if it never overestimates the cost of the best
+ * reachable goal node, i.e. c(n*) <= c(n) + h(n) for cost function c,
+ * heuristic h, any node n in the search graph, and n* the best reachable goal
+ * node from n.
+ */
+[[maybe_unused]] static inline bool isAdmissible(const Heuristic heuristic) {
```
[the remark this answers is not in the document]
EliasLF: The mapper itself currently only needs to know if the heuristic is principally admissible. `isAdmissible` is therefore only used in the tests (which would result in a compiler warning, if the project is compiled without the tests). I thought about removing the function, but came to the conclusion that it's not unlikely that we come up with optimizations specifically for generally admissible heuristics further down the line. So its probably best to keep that information on all our heuristics, especially since its basically just 1 switch statement.
Also it is just a great sanity check in the tests. I caught many bugs in the current implementation only because I started to test for general admissibility.

include/configuration/Heuristic.hpp
```diff
@@ -0,0 +1,158 @@
...
+  switch (heuristic) {
+  case Heuristic::GateCountMaxDistance:
+  case Heuristic::FidelityBestLocation:
+    return true;
+  case Heuristic::GateCountSumDistanceMinusSharedSwaps:
+  case Heuristic::GateCountMaxDistanceOrSumDistanceMinusSharedSwaps:
+  case Heuristic::GateCountSumDistance:
+    return false;
+  }
+  return false;
```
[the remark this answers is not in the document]
EliasLF: Hm, that's true. I just wonder if it might be better in this specific case to stay explicit. This way, linter warnings for missing switch cases will remind us to keep tracking those properties for new heuristics.

include/heuristic/HeuristicMapper.hpp
EliasLF: That's right. I just followed the pre-existing lines above without much thought. I'm not too familiar with how C++ handles struct constructors in this regard, but don't we save an unnecessary vector allocation this way, since the copy constructor would overwrite the default initialization? Or does the default initialization only happen if there is no other initialization in the constructor?

include/heuristic/HeuristicMapper.hpp
EliasLF: Isn't an enum much more efficient in many regards (space-wise and runtime-wise in comparisons), since it's only an int in the backend?

burgholzer:
> Alright, first of all thank you very much for the quick review!
> 
> Some of the design choices in this PR were made with future plans/issues in mind, which are (or were previously) not mentioned in the PR description above. Before going into detail in your code comments, maybe let's first discuss them in general here:

Thanks for the detailed answer. This cleared up quite a lot. I'll also start here before going in-depth with the comments.
 
> * semi-directional architectures (i.e. architectures with both bi- and unidirectional edges)
> * user-defined gate costs (for the non-fidelity-aware case)
> * arbitrary 2q gates
> 
> If I'm not mistaken, we already talked about all of these in person but it's probably a good idea to also get this into writing here on Github:
> 
> The order above is (most likely) also the order of their relevancy (from low to high) and unfortunately also the order of the incurred cost/complexity (from low to high).
> 
> To my knowledge currently there exist no semi-directional architectures in practice. However, our heuristic mapper is implemented abstractly enough, that allowing for such architectures comes almost for free (2 extra jump instructions per search node (for purely bi-/unidirectional architectures), a boolean field `Architecture::isUnidirectional`, a getter method for that field and 1 small else-case in the loop of `Architecture::createDistanceTable`). Originally, the plan was to tackle this issue in a future PR, but thanks to your comments I just realized that there were only 2 lines remaining (in the heuristic mapper) that assumed pure directionality. I therefore just pushed these remaining changes (+ a few optimizations) and added this "feature" to the PR description. If you disagree with the decision to allow semi-directional architectures, it would be quite easy to revert the changes. In that case, however, I think we should at least specifically check and disallow loading such coupling maps (which currently is not the case, they are just not correctly handled during mapping)

Now that makes a lot more sense! Although such architecture do not exist in practice at the moment, it could very much happen and it almost never hurts to be a little more general. So I am happy with the changes here! Even better to hear, that the review helped in identifying the remaining places that needed changes.

> User-defined gate costs and arbitrary 2q gates are probably much more relevant, but also more complex, which I only realized recently. 

I have feared as much. Although I still hope that the number of cases that need to be added to handle arbitrary two-qubit gates stays reasonably low. In fact, for bidirectional architectures there shouldn't be too many changes at all. There might be some more optimization potential (e.g., commutation rules involving controlled gates), but we are not taking advantage of that at the moment anyway. The unidirectional case might be trickier, but I think the existing PR already laid a solid foundation for that.

> As mentioned in the PR description, even the current implementation strictly speaking wrongly assumes a convex cost space due to cumulative reversal costs on logical edges possibly surpassing swap costs (in the non-fidelity-aware case on non-bidirectional architectures). E.g. if there are 9 congruent 2q-gates in a layer, which are already validly mapped to a back-edge incurring 9_4=36 in reversal costs, swapping them to a neighboring forward-edge with 1 swap only costs 34, i.e. the first goal node on the search path is not the optimal goal node. This is currently not a huge problem, since it's rare to find 9 congruent CNOTs in 1 layer (and even only possible with `Disjoint2qBlock` layering) and could easily be optimized down to at most 3 CNOTs in the pre-optimization stage. However, with more evened out costs between H and CNOT, or different reversal operations for other 2q gates, this non-convexity might become more significant. The new distance/dijkstra system (not only fixes the currently flawed, non-admissible system, but also) gives flexibility for future heuristics tackling this non-convexity problem by dropping the tightness constraint. In contrast to the previous custom dijkstra solution, edge-skipping dijkstra will return the correct distance for any combination of swap cost and reversal cost; and 1 execution with 1 skip (as it is currently used in the non-fidelity-aware case) it only increases the runtime complexity from O(E_log V) to O(E_log V + V²_E), which is not too bad for something that runs only once per mapping.

I agree that this mostly seems like a theoretical issue for now. Especially since I would guess that the cost difference between single- and two-qubit gates will always stay rather big.
I also agree that the runtime trade-off is definitely worth it. Technically it is not once per mapping, but once per architecture. That information could be pre-computed and re-used, right? At some point we might need a system for such pre-computations (Similar how we did that for the sub-architectures feature; that is unfortunately not as neatly integrated as I would love it to be)

include/Architecture.hpp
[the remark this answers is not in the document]
burgholzer: I see now. Thanks for the explanation. I think a comment should be enough to clear this up.
Could it help to rename these to something like `isPurelyBidirectional` and `isPurelyUnidirectional` or maybe `isEntirelyBidirectional`, etc.?

include/Architecture.hpp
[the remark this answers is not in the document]
burgholzer: Alright. With the added context you provided, things become much clearer.
The added flexibility certainly is a plus here. And as you said: it might be necessary for handling arbitrary two-qubit gates.
Space-wise, it's also not really an issue; even if we were to consider 1000+ qubit architectures.
Runtime-wise it's a bit more of an overhead, but I guess it's also fine. (Again, some kind of caching could help in addition)

include/configuration/Heuristic.hpp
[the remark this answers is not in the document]
burgholzer: I see. Makes sense and I am totally on board with keeping the function 👍🏼

include/configuration/Heuristic.hpp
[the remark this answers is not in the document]
burgholzer: Good point 👍🏼 feel free to resolve this then!

include/heuristic/HeuristicMapper.hpp
burgholzer: I am fairly sure that you are right when no optimizations are involved, but I think that modern compilers are clever enough to optimize any unnecessary initialisations away.
I guess you could just replace the existing code and see if anything breaks 🤷🏼

include/heuristic/HeuristicMapper.hpp
burgholzer: That kind of depends I guess. That Heuristics class would probably not have all that many member. So its memory footprint would probably be comparable. The comparison logic kind of depends on how it's implemented in the class, but could also be pretty comparable.
However, I am not really trying to push towards such a solution. I am not yet sure myself if it would amount to something useful. Let's keep it as it is for now, but keep the additional option in mind ☺️

include/Architecture.hpp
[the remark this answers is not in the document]
EliasLF: I'm definitely open to renaming them.
However, in my opinion it is not strictly necessary now that there are doc strings. I think it mostly just adds clutter. And generally I prefer the semantics of calling the property of a set, which only holds if some property holds for all elements of the set, by the exactly same name as the properties of the elements (the same way e.g. a system of equations is called "linear" and not "purely linear"), but I see how that is purely subjective.

include/heuristic/HeuristicMapper.hpp
EliasLF: Alright, sure, let's keep it in mind☺️

include/Architecture.hpp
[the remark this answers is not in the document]
burgholzer: I agree with that sentiment! 👍🏼 
let's keep the names as they are!

include/heuristic/HeuristicMapper.hpp
EliasLF: Alright, I looked into this in more detail:

For vector members:
1. A simple copy assignment in the constructor body, does actually result in 2 initializations (the default initialization and a reinitialization), which also seems to not get optimized away by GCC (looking at the assembly of a minimal example)
2. By using the initialization list of the constructor the default initialization is skipped and the vector is directly initialized to the correct size, resulting in only 1 initialization.
3. `std::copy` of course also results in 1 initialization (the default initialization), but the vector might get resized during copying

For array members all 3 methods are equivalent copying just the data over, since the arrays are allocated within the struct.

So in theory 2. (initialization list) should be the fastest, then 3. (`std::copy`) and then 1. (copy assignment)

I ran a few tests with QMAP comparing the runtime (averaging multiple runs, which were executed in parallel with varying starting order to reduce effects of environment and scheduler).
I have _no_ idea why, but the order of their mapping times is exactly reversed to what I expected. The initialization list is about 1% slower than the current solution, and copy assignments in the constructor body are about 0.5% faster. These differences aren't huge but I ran the experiment often enough, that the results are statistically significant.

So should we switch to copy assignments (in that case we probably need to disable the linter warning to use the initialization list on those lines) or leave it as is? Or do we take the 1% performance decrease for slightly cleaner code?


Btw, while at it, I also tested the effects of using vectors instead of arrays for `qubits` and `locations` (both in the mapper and the nodes; using fixed size vectors with equal length to the number of output qubits), since the current solution of hard-coded 128-length arrays is both pretty wasteful space-wise and not too future-proof. Unfortunately, this results in a significant performance decrease of about 40%.

include/heuristic/HeuristicMapper.hpp
EliasLF: That's true, thank you!

test/test_heuristic.cpp
EliasLF: Hm true, I could have sworn I had been running into compiler errors/warnings without the explicit conversion, but it seems to work now. Great, thanks!

src/heuristic/HeuristicMapper.cpp
EliasLF: Most of this probably cleared up with the discussion of semi-directionality. The use of both ` unidirectional()` and `bidirectional()` is intentional to account for semi-directional architectures.

This heuristic relies on subtracting the maximum cost savable up to the optimal solution via shared swaps.
So when calculating the number of swaps in each distance, we need to assume the cheapest possible swap cost for the highest possible number of swaps (which only in the case of unidirectional architectures is the higher `COST_UNIDIRECTIONAL_SWAP`)
Later we subtract the highest cost resulting from a given number of shared swaps, where we need to assume the _highest_ possible cost per swap (which only for bidirectional architectures is the lower `COST_BIDIRECTIONAL_SWAP`)

So in one case semi-directional architectures are handled together with bidirectional ones (since that is the worst case), and in another there are handled together with unidirectional architectures (since that is the worst case there)

src/Architecture.cpp
EliasLF: Actually, now that you mention it, this is the case for `buildEdgeSkipTable`, which could directly call `buildTable` for its first dimension. I will quickly implement that and push it to this PR.

However, in the above case with `buildSingleEdgeSkipTable` we potentially need to build multiple different edge skip tables with different reversal costs. Currently, it is just 1 or 2 tables, which could just be communicated to the function via parameters, though even this would already unnecessarily increase coupling between those 2 classes in my opinion.
However, going forward with arbitrary 2q gates it will quite likely become necessary to build edge skip tables for even more different reversal costs and potentially do that even on the fly later on. So I think it is better to keep `buildTable` and `buildSingleEdgeSkipTable` separate from each other.

src/configuration/Configuration.cpp
EliasLF: Great idea, thank you!

src/heuristic/HeuristicMapper.cpp
EliasLF: Great idea, I just implemented that and pushed it to this PR.

EliasLF:
> Technically it is not once per mapping, but once per architecture. That information could be pre-computed and re-used, right? At some point we might need a system for such pre-computations

I agree, pre-computing all the distance values for common architectures (and maybe even a system for saving them to a file for custom architectures) is a great idea and would open the possibility of using metrics that are otherwise too expensive to compute for each mapping process.
For example by computing _all_ possible distances for _any_ reversal cost (which are finitely many, since the cheapest path can only change so many times before skipping the most expensive edge) we could solve the reversal cost problem robustly for the general case.

Too bad, the same is not possible/useful for fidelity distances because of their variability over time. Pre-computing viable swap sharing combinations would solve so much of the gap between the heuristic and true cost there.

include/heuristic/HeuristicMapper.hpp
burgholzer: Thanks for the thorough testing. 
I guess I would just take the 1% performance hit to have the cleaner code. That 1% won't really be noticeable in real world examples.

It's really unfortunate that the change from arrays to vectors is so costly. But maybe we'll find a way to circumvent that somehow in the future.

src/Architecture.cpp
burgholzer: Alright. Let's keep them separate for now. I guess the 2q-gates feature will show if this can be simplified or becomes even more complicated.

src/heuristic/HeuristicMapper.cpp
burgholzer: Yeah. That is all clear now. And now that o understand it better, it is quite elegant 😌

burgholzer:
> > Technically it is not once per mapping, but once per architecture. That information could be pre-computed and re-used, right? At some point we might need a system for such pre-computations
> 
> 
> 
> I agree, pre-computing all the distance values for common architectures (and maybe even a system for saving them to a file for custom architectures) is a great idea and would open the possibility of using metrics that are otherwise too expensive to compute for each mapping process.
> 
> For example by computing _all_ possible distances for _any_ reversal cost (which are finitely many, since the cheapest path can only change so many times before skipping the most expensive edge) we could solve the reversal cost problem robustly for the general case.
> 
> 
> 
> Too bad, the same is not possible/useful for fidelity distances because of their variability over time. Pre-computing viable swap sharing combinations would solve so much of the gap between the heuristic and true cost there.

Although fidelity data varies over time, it all depends on the frequency of calibrations whether it is worth to invest the time and pre-compute these values.
If calibration only happens once every couple hours and computing the tables takes a couple minutes (assuming HPC resources available), it could be worth it.

src/heuristic/HeuristicMapper.cpp
EliasLF: Thanks😊

EliasLF:
> Although fidelity data varies over time, it all depends on the frequency of calibrations whether it is worth to invest the time and pre-compute these values.
> If calibration only happens once every couple hours and computing the tables takes a couple minutes (assuming HPC resources available), it could be worth it.

Hm, interesting point. I guess if you would map multiple circuits per calibration cycle, that's true, yes. I will keep it as an option in mind then👍

PR #411, opened 2024-01-05
Fix `generatedNodes` benchmark bug

EliasLF:
## Description

Fixing a minor bug, where sometimes the number of generated nodes is undercounted in the benchmark because nodes with equivalent layout are not added multiple times to the priority queue. Therefore, instead of relying on the size of the priority queue, it is more stable to rely on the node ids, which are strictly incremented for each generated node.

## Checklist:


burgholzer (approved):
LGTM

PR #416, opened 2024-01-19
Fidelity-aware lookahead, lookahead-based initial layout, `DisjointSameOpType` layering, upper limits for active qubits per layer

EliasLF:
## Description

- introducing 2 fidelity-aware lookahead heuristics (`FidelityCurrentLocation` just sums up the fidelity costs and distances of/within future gates with the current mapping, `FidelityCurrentTargetLocation` does the same but for the mapping the heuristic is currently aiming for, e.g. for `FidelityBestLocation` this is the globally, individual best location for each qubit)
- introducing a new initial layouting technique `DynamicGreedyLookahead`, which is similar to `Dynamic` but instead of just looking at distances in the current layer, each qubit is mapped greedily such that the lookahead penalty is minimized. This increases the quality of mapping results significantly (also in the non-fidelity-aware case) in almost all cases with not too much runtime overhead. We might consider making this the new default.
- A new layering technique `DisjointSameOpType` equivalent to `Disjoint2qBlocks` but separating different gate types to different layers. This is necessary for fidelity-aware mapping, where `Disjoint2qBlocks` might make it impossible to find the global optimum.
- Introducing user-configurable upper limits for active qubits per layer to reduce mapping complexity in return for less optimal mapping results.
- Generally refactored the layering process slightly to make future additions simpler and each layering procedure more atomic.

## Checklist:

- [ ] The pull request only contains commits that are related to it.
- [ ] I have added appropriate tests and documentation.


burgholzer:
@EliasLF: just checking in; has there been any progress on this?

PR #454, opened 2024-05-16
🚸 Hide `uninstall` target when it exists in superproject

ystade:
## Description

When QMap is included in another project that also provides an uninstall target, this leads to a conflict. This resolves this issue by creating this target only when there does not exist already one and fixes some bugs with the Z3 import.

## Checklist:


src/CMakeLists.txt
burgholzer, line 1: Is this really the best way to go forward here? I think by now we simply require Z3 to be installed.

burgholzer (approved):
LGTM in principle. Thanks. Just an inline question on the Z3 related changes.

src/CMakeLists.txt
ystade: How you wish, I do not have any strong opinion here. However, then one should definitely also adapt the lines
https://github.com/cda-tum/mqt-qmap/blob/94ee532c53beb5b33823445d752b0356956d629f/cmake/ExternalDependencies.cmake#L8-L11

burgholzer: Nah, nevermind. But could you please just move the alias definitions for the exact mapper and the Clifford synthesis library into the if block just a couple lines above the newly created one? Makes much more sense to me to have these right after the corresponding add_qmap_library calls anyways.

ystade: Done ✅

burgholzer (approved):
Many thanks.
The test failures are due to Qiskit 1.1 being released overnight and introducing new warning.
The respective warning here should be ignored in our pyproject.toml
Do you know how to do that or should I give it a try later on?

ystade:
> Many thanks. The test failures are due to Qiskit 1.1 being released overnight and introducing new warning. The respective warning here should be ignored in our pyproject.toml Do you know how to do that or should I give it a try later on?

No, I have no clue how... :/

burgholzer:
> > Many thanks. The test failures are due to Qiskit 1.1 being released overnight and introducing new warning. The respective warning here should be ignored in our pyproject.toml Do you know how to do that or should I give it a try later on?
> 
> No, I have no clue how... :/

https://github.com/cda-tum/mqt-qmap/pull/454/commits/70c854854def3daf4263db3ceb05436a126573ad should do the trick.
The syntax for this is available under the following link if you should ever need it: https://docs.python.org/3/library/warnings.html#describing-warning-filters

PR #458, opened 2024-05-24
Hybrid Neutral Atom Mapper

lsschmid:
## Description

Adds the hybrid mapping functionality for Neutral Atom Hardware.

## Checklist:


include/hybridmap/HybridNeutralAtomMapper.hpp
```diff
@@ -0,0 +1,534 @@
...
+#include "hybridmap/NeutralAtomArchitecture.hpp"
+#include "hybridmap/NeutralAtomDefinitions.hpp"
+#include "hybridmap/NeutralAtomScheduler.hpp"
+#include "hybridmap/NeutralAtomUtils.hpp"
+#include "map"
+#include "operations/Operation.hpp"
+#include "set"
+#include "string"
+#include "utility"
+#include "vector"
```
burgholzer, line 25:
```suggestion
#include "Definitions.hpp"
#include "NeutralAtomLayer.hpp"
#include "QuantumComputation.hpp"
#include <cstdint>
#include <deque>
#include <fstream>
#include "hybridmap/HardwareQubits.hpp"
#include "hybridmap/Mapping.hpp"
#include "hybridmap/NeutralAtomArchitecture.hpp"
#include "hybridmap/NeutralAtomDefinitions.hpp"
#include "hybridmap/NeutralAtomScheduler.hpp"
#include "hybridmap/NeutralAtomUtils.hpp"
#include <map>
#include "operations/Operation.hpp"
#include <set>
#include <string>
#include <utility>
#include <vector>
```

include/hybridmap/HybridNeutralAtomMapper.hpp
```diff
@@ -0,0 +1,534 @@
...
+struct MultiQubitMovePos {
+  CoordIndices coords;
+  size_t       nMoves{0};
+
+  // copy constructor
+  MultiQubitMovePos()                               = default;
+  MultiQubitMovePos(const MultiQubitMovePos& other) = default;
+
+  // copy assignment
+  MultiQubitMovePos& operator=(const MultiQubitMovePos& other) = default;
```
burgholzer, line 58: Not 100% sure, but very certain that you can just remove all of these (copy) constructors here and the code will still work. Instances can then simply be constructed using initializer lists (`{...}`)

include/hybridmap/HybridNeutralAtomMapper.hpp
```diff
@@ -0,0 +1,534 @@
...
+  fp       gateWeight           = 1;
+  fp       shuttlingWeight      = 1;
+  uint32_t seed                 = 0;
+};
+
+/**
+ * @brief Helper struct to store the position of a multi qubit gate and the
+ * number of moves needed to execute it.
+ */
+struct MultiQubitMovePos {
```
burgholzer, line 49: Seems like a utility class to me. Might be worth moving to the Utils file

include/hybridmap/HybridNeutralAtomMapper.hpp
```diff
@@ -0,0 +1,534 @@
...
+  // Counter variables
+  uint32_t nSwaps = 0;
+  uint32_t nMoves = 0;
+
+  // The current placement of the hardware qubits onto the coordinates
+  HardwareQubits hardwareQubits;
+  // The current mapping between circuit qubits and hardware qubits
+  Mapping mapping;
+
+  bool verbose = false;
```
burgholzer, line 118: Shouldn't that also be a MapperParameter?

include/hybridmap/HybridNeutralAtomMapper.hpp
```diff
@@ -0,0 +1,534 @@
...
+  /**
+   * @brief Calculates the cost of a series of move operations by summing up the
+   * cost of each move.
+   * @param moveComb The series of move operations to compute the cost for
+   * @return The total cost of the series of move operations
+   */
+  fp moveCostComb(const MoveComb& moveComb);
+
+  // temp
+  void printLayers();
```
burgholzer, line 367: temp? worth fixing?

include/hybridmap/HybridNeutralAtomMapper.hpp
```diff
@@ -0,0 +1,534 @@
...
+
+public:
+  // Constructors
+  [[maybe_unused]] NeutralAtomMapper(const NeutralAtomMapper&) = delete;
```
burgholzer, line 371: Why explicitly delete the copy constructor?

include/hybridmap/HybridNeutralAtomMapper.hpp
```diff
@@ -0,0 +1,534 @@
...
+   * @param qc  The quantum circuit to be mapped
+   * @param initialMapping The initial mapping of the circuit qubits to the
+   * hardware qubits
+   */
+  [[maybe_unused]] void mapPy(qc::QuantumComputation& qc,
+                              InitialMapping initialMapping, bool printInfo) {
+    map(qc, initialMapping, verbose = printInfo);
+    convertToAod(this->mappedQc);
+  }
+
```
burgholzer, line 439: If only used for the python bindings, then consider moving this function there. Both of the methods used are public methods.

include/hybridmap/HybridNeutralAtomMapper.hpp
```diff
@@ -0,0 +1,534 @@
...
+   * @return The results of the scheduler
+   */
+  [[maybe_unused]] SchedulerResults schedule(bool verboseArg           = false,
+                                             bool createAnimationCsv   = false,
+                                             fp   shuttlingSpeedFactor = 1.0) {
+    scheduler = NeutralAtomScheduler(arch);
+    return scheduler.schedule(mappedQcAOD, hardwareQubits.getInitialHwPos(),
+                              verboseArg, createAnimationCsv,
+                              shuttlingSpeedFactor);
+  }
```
burgholzer, line 499: Should scheduling really be part of the mapper here?
wouldn't it make sense to decouple this and have a mapping stage that is followed by a separate scheduling stage?

include/hybridmap/HybridNeutralAtomMapper.hpp
```diff
@@ -0,0 +1,534 @@
...
+   * @param qc The already mapped quantum circuit with abstract SWAP gates and
+   * MOVE operations
+   * @return The mapped quantum circuit with native AOD operations
+   */
+  QuantumComputation convertToAod(qc::QuantumComputation& qc);
+
+  [[maybe_unused]] [[nodiscard]] std::map<HwQubit, HwQubit>
+  getInitHwPos() const {
+    return hardwareQubits.getInitialHwPos();
+  }
```
burgholzer, line 531: if unused, this can also be removed IMHO

include/hybridmap/HybridNeutralAtomMapper.hpp
```diff
@@ -0,0 +1,534 @@
...
+    return scheduler.getAnimationCsv();
+  }
+
+  /**
+   * @brief Saves the animation csv file of the scheduled quantum circuit.
+   * @param filename The name of the file to save the animation csv file to
+   */
+  [[maybe_unused]] void saveAnimationCsv(const std::string& filename) {
+    scheduler.saveAnimationCsv(filename);
+  }
```
burgholzer, line 515: GIven the comment above, these could be removed.

include/hybridmap/AodScheduler.hpp
```diff
@@ -0,0 +1,302 @@
...
+        }
+      }
+
+      [[nodiscard]] std::vector<AodMove*> getActivates(Dimension dim) const {
+        if (dim == Dimension::X) {
+          return activateXs;
+        }
+        return activateYs;
+      }
+    };
```
burgholzer, line 88: The use of `new` here without corresponding `delete` calls screams memory leaks. Consider using smart pointers (probably `std::shared_ptr`) instead or to replace the pointers by pure values in general (assuming the pointer semantics are not needed).

include/hybridmap/AodScheduler.hpp
```diff
@@ -0,0 +1,302 @@
...
+
+    // AODScheduler
+    // NeutralAtomArchitecture to call necessary hardware information
+    NeutralAtomArchitecture    arch;
```
burgholzer, line 92: Another copy of the architecture? Might be an architectural decision to rethink. Probably the easiest fix is to replace this with a `NeutralAtomArchitecture*` and use `&arch` in the constructor.

include/hybridmap/AodScheduler.hpp
```diff
@@ -0,0 +1,302 @@
...
+     * @param opsFinal Unloading operations
+     * @return The shuttling operation between the loading and unloading
+     * operations
+     */
+    static AodOperation
+    connectAodOperations(const std::vector<AodOperation>& opsInit,
+                         const std::vector<AodOperation>& opsFinal);
+  };
+
+  NeutralAtomArchitecture arch;
```
burgholzer, line 265: Yet another copy. See comments above

include/hybridmap/NeutralAtomScheduler.hpp
```diff
@@ -0,0 +1,160 @@
...
+  std::string        getEndString(fp endTime);
+  static std::string createCsvLine(fp startTime, HwQubit id, fp x, fp y,
+                                   uint32_t size = 1, uint32_t color = 0,
+                                   bool axes = false, axesId axId = 0,
+                                   bool margin = false, marginId marginId = 0,
+                                   fp marginSize = 0);
+  std::string createCsvOp(const std::unique_ptr<Operation>& op, fp startTime,
+                          fp endTime, const qc::NeutralAtomArchitecture& arch);
+};
+
```
burgholzer, line 159: It might be worth to separate the animation code from the actual library code just to get a better separation of concerns. Maybe put all the related code into an `Animations.hpp` file.

src/hybridmap/AodScheduler.cpp
```diff
@@ -0,0 +1,549 @@
...
+                         [&moveToRemove](const auto& movePair) {
+                           return movePair.first == moveToRemove;
+                         }),
+          groupIt->moves.end());
+    }
+    if (!possibleNewMoveGroup.moves.empty()) {
+      groupIt = moveGroups.insert(groupIt + 1, std::move(possibleNewMoveGroup));
+      possibleNewMoveGroup = MoveGroup{arch};
+      groupIt--;
+    }
```
burgholzer, line 315: This code path isn't covered by the existing tests. Is that due to the nature of the tests or is this essentially dead code that could be removed?

include/hybridmap/AodScheduler.hpp
burgholzer, line 1: It might be worth to consolidate the two components that are called `Scheduler` here, if possible.

src/hybridmap/HybridNeutralAtomMapper.cpp
```diff
@@ -0,0 +1,1293 @@
...
+#include "Definitions.hpp"
+#include "QuantumComputation.hpp"
+#include "hybridmap/AodScheduler.hpp"
+#include "hybridmap/NeutralAtomDefinitions.hpp"
+#include "hybridmap/NeutralAtomLayer.hpp"
+#include "hybridmap/NeutralAtomUtils.hpp"
+#include "iterator"
+#include "operations/OpType.hpp"
+#include "operations/Operation.hpp"
+#include "utils.hpp"
```
burgholzer, line 18:
```suggestion
#include "CircuitOptimizer.hpp"
#include "Definitions.hpp"
#include "QuantumComputation.hpp"
#include "hybridmap/AodScheduler.hpp"
#include "hybridmap/NeutralAtomDefinitions.hpp"
#include "hybridmap/NeutralAtomLayer.hpp"
#include "hybridmap/NeutralAtomUtils.hpp"
#include <iterator>
#include "operations/OpType.hpp"
#include "operations/Operation.hpp"
#include "utils.hpp"
```

src/hybridmap/HybridNeutralAtomMapper.cpp
```diff
@@ -0,0 +1,1293 @@
...
+    }
+  }
+}
+
+void qc::NeutralAtomMapper::mapGate(const Operation* op) {
+  if (op->getType() == qc::OpType::I) {
+    return;
+  }
+  // Safety check
+  // TODO remove
```
burgholzer, line 187: Can this todo be completed?

src/hybridmap/HybridNeutralAtomMapper.cpp
```diff
@@ -0,0 +1,1293 @@
...
+                                       const WeightedSwaps& swapExact) {
+  fp distBefore = 0;
+  fp distAfter  = 0;
+  fp distChange = 0;
+  // bring close only until swap distance =0, bring exact to the exact position
+  // bring qubits together to execute gate
+  for (const auto& [q1, q2] : swapCloseBy) {
+    // distance before
+    distBefore = this->hardwareQubits.getSwapDistance(q1, q2);
+    if (distBefore == std::numeric_limits<fp>::infinity()) {
```
burgholzer, line 461: Stated this elsewhere already, but just to be sure: prefer to use finite values, e.g. `max()`. Please also check throughout this file

src/hybridmap/HybridNeutralAtomMapper.cpp
```diff
@@ -0,0 +1,1293 @@
...
+    this->frontLayerShuttling.push_back(op);
+  }
+  // remove from lookahead layer if there
+  auto idxLookaheadGate = std::find(this->lookaheadLayerGate.begin(),
+                                    this->lookaheadLayerGate.end(), op);
+  if (idxLookaheadGate != this->lookaheadLayerGate.end()) {
+    this->lookaheadLayerGate.erase(idxLookaheadGate);
+    this->lookaheadLayerShuttling.push_back(op);
+  }
+  return {};
```
burgholzer, line 586: All of this code is uncovered by tests. Is it actually needed? if yes, can it be tested?

src/hybridmap/HybridNeutralAtomMapper.cpp
```diff
@@ -0,0 +1,1293 @@
...
+        }
+        // remove from lookahead layer if there
+        auto idxLookaheadGate = std::find(this->lookaheadLayerGate.begin(),
+                                          this->lookaheadLayerGate.end(), op);
+        if (idxLookaheadGate != this->lookaheadLayerGate.end()) {
+          this->lookaheadLayerGate.erase(idxLookaheadGate);
+          this->lookaheadLayerShuttling.push_back(op);
+        }
+        return {};
+      }
```
burgholzer, line 701: Same question as above: currenlty untested. testable or removable?

include/hybridmap/AodScheduler.hpp
lsschmid: They are actually two different things. The AodScheduler converts the Move Operations into the native gate set (AodActivate/Dacative and AodMove). The NeutralAtomScheduler does not convert anything but just schedules all operations to compute the overall time. I might consider renaming the AodScheduler to AodConverter or sth similer

include/hybridmap/AodScheduler.hpp
lsschmid: Tried to not use pointers at all. Failed. Tried again. Failed again. Now I switched to shared_ptr

include/hybridmap/AodScheduler.hpp
lsschmid: This is indeed not ideal, but the AodScheduler needs some information on the atom spacing which is provided by the Architecture class.
Regarding the pointer, I thought that the fact that I use std::move(arch) should avoid an additional copy.

include/hybridmap/HardwareQubits.hpp
```diff
@@ -0,0 +1,290 @@
...
+        seed = std::random_device()();
+      }
+      std::mt19937 g(seed);
+      std::shuffle(indices.begin(), indices.end(), g);
+      for (uint32_t i = 0; i < arch.getNqubits(); ++i) {
+        hwToCoordIdx.insert({i, indices[i]});
+      }
+
+      swapDistances = SymmetricMatrix(arch.getNqubits(), -1);
+    }
```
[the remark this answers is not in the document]
lsschmid: This is actually a bug in the test, fixed it there.

include/hybridmap/AodScheduler.hpp
burgholzer: Unfortunately, the initialization of the member will still create a copy. In case of passing by reference, the copy happens in the constructor itself. In case of the move version, the architecture itself is passed by value, hence, creating a copy.

But I believe the solution here is rather straightforward by simply replacing the Architecture member with a pointer to an architecture. Happy to try that out tomorrow, if you don't feel comfortable with it.

include/hybridmap/AodScheduler.hpp
burgholzer: Yeah. I figured they are not really doing the same thing. A rename sounds reasonable. Like your suggestion. Could be even more explicit with something like `MoveToAodConverter`

include/hybridmap/AodScheduler.hpp
burgholzer: I can also give this a try tomorrow and see whether I can get it to work without pointers 😌

include/hybridmap/HardwareQubits.hpp
```diff
@@ -0,0 +1,290 @@
...
+  void move(HwQubit hwQubit, CoordIndex newCoord);
+
+  /**
+   * @brief Converts gate qubits from hardware qubits to coordinate indices.
+   * @param op The operation.
+   */
+  void mapToCoordIdx(Operation* op) const {
+    op->setTargets(hwToCoordIdx.apply(op->getTargets()));
+    op->setControls(hwToCoordIdx.apply(op->getControls()));
+  }
```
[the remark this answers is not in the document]
lsschmid: Until now I removed all measurements, so this was not a problem. I generalized the parts to also support non-unitary operations

include/hybridmap/AodScheduler.hpp
lsschmid: Ah, then I had a wrong understanding of the passing by reference. Can then also change to the pointer version, no problem

include/hybridmap/NeutralAtomLayer.hpp
```diff
@@ -0,0 +1,98 @@
...
+   * @param gatesToRemove Gates to remove from the current layer
+   * @param commuteWith Gates the new gates should commute with
+   */
+  void removeGatesAndUpdate(const GateList& gatesToRemove);
+  /**
+   * @brief Returns the mapped single qubit gates
+   * @return The mapped single qubit gates
+   */
+  GateList getMappedSingleQubitGates() { return mappedSingleQubitGates; }
+};
```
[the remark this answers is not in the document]
lsschmid: Regarding the layers there actually exist differences, as Yannick considered global single qubit rotations, so one actually would need to check if/how one could combine the two. Regarding commutation rules, this actually would make sense I guess.

include/hybridmap/NeutralAtomUtils.hpp
```diff
@@ -0,0 +1,276 @@
...
+};
+
+/**
+ * @brief Helper class to manage multiple atom moves which belong together.
+ * @details E.g. a move-away combined with the actual move. These are combined
+ * in a MoveComb to facilitate the cost calculation.
+ */
+struct MoveComb {
+  std::vector<AtomMove> moves;
+  fp                    cost = std::numeric_limits<fp>::quiet_NaN();
```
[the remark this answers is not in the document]
lsschmid: I always thought NaN is the the better option, as it will throw an error if used in the wrong way, while max numbers might stay unnoticed.

src/hybridmap/HardwareQubits.cpp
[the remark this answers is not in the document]
lsschmid: Good point. The original idea was that the architecture class should only provide "unchangeable" hardware information which is specific to the hardware. HardwareQubits manage the the "state of the hardware" during mapping. So the idea is that from one mapping run to the other, the architecture stays exactly the same but a new "set" of hardware qubits is used, if that makes sense.

src/hybridmap/Mapping.cpp
```diff
@@ -0,0 +1,30 @@
...
+  auto q2 = swap.second;
+  if (this->isMapped(q1) && this->isMapped(q2)) {
+    auto circQ1 = this->getCircQubit(q1);
+    auto circQ2 = this->getCircQubit(q2);
+    this->setCircuitQubit(circQ2, q1);
+    this->setCircuitQubit(circQ1, q2);
+  } else if (this->isMapped(q1) && !this->isMapped(q2)) {
+    this->setCircuitQubit(this->getCircQubit(q1), q2);
+  } else if (this->isMapped(q2) && !this->isMapped(q1)) {
+    this->setCircuitQubit(this->getCircQubit(q2), q1);
```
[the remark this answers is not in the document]
lsschmid: I know but usually I try to use it, as I find it helpful when reading the code.

include/hybridmap/NeutralAtomUtils.hpp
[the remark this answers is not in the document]
burgholzer: In principle, that is true. However, on the one hand, you are using `quiet` NaNs, which IIRC do not really throw errors. So in many cases, they will remain unnoticed and choas will arise elsewhere.
On the other hand, all these non-finite special numbers need special handling that is computationally non-trivial. We explicitly enable certain compiler optimizations that assume finite computations (`-ffinite-math`) that generally improve performance, but where one should avoid the non-finite numbers.
If you truly want to represent an uninitialized value, you might want to look at `optional<...>`. But I think this is not really necessary here.

src/hybridmap/HardwareQubits.cpp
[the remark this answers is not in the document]
burgholzer: Yeah. That kind of makes sense. I think it could still be an idea to combine both, as they are heavily related. The architecture could then have a `resetState` method (or similar) that resets the state back. This `State` could be an internal class of the architecture.

test/hybridmap/test_hybridmap.cpp
```diff
@@ -0,0 +1,154 @@
...
+  auto qcMapped = mapper.map(qc, initialMapping, true);
+
+  std::ofstream dummyFs;
+  qcMapped.dumpOpenQASM(dummyFs, false);
+
+  auto qcAodMapped = mapper.convertToAod(qcMapped);
+  qcAodMapped.dumpOpenQASM(dummyFs, false);
+
+  auto scheduleResults = mapper.schedule(true, true);
+  dummyFs << scheduleResults.toCsv();
```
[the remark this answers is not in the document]
lsschmid: I wasn't sure how to do it, as dumpOpenQasm requires a filestream object. But if it's possible apparently, I'll add it 👍

src/hybridmap/Mapping.cpp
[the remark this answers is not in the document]
burgholzer: I guess that's a matter of personal preference. I see where you are coming from. To me, it mostly feels like a waste of 6 characters that lengthens the code without adding that much value.
Given that we use object oriented programming and there are hardly any free floating functions, any function call can almost always only be a member function call. So the `this` truly does not add that much value in my opinion. There is simply almost no other option where a function could be coming from.
This shouldn't be a veto against adding it though. Just wanted to provide a little more context on why I personally do not add it to our code bases.

test/python/test_hybridmap.py
```diff
@@ -0,0 +1,94 @@
...
+            "cccz": 0.95,
+            "ccccz": 0.95,
+            "cccccz": 0.95,
+            "mcx": 0.95,
+        },
+        "decoherenceTimes": {"t1": 100000000, "t2": 1500000},
+        "shuttlingTimes": {"move": 0.55, "aod_move": 0.55, "aod_activate": 20, "aod_deactivate": 20},
+        "shuttlingAverageFidelities": {"move": 1, "aod_move": 1, "aod_activate": 1, "aod_deactivate": 1},
+    },
+}
```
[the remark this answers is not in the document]
lsschmid: I actually had the parametrized tests and used the files based on __file__. But I couldn't make it run on the CI (maybe you can check some commits in the past). I also tested the creation of temporary files but also couldn't really make it work, so this is the workaround which works the best.

include/hybridmap/NeutralAtomLayer.hpp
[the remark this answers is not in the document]
burgholzer: Agreed that there are subtle differences in the layers and how they are formed. 
The general concept still is the same, and the original QMAP code for example also has multiple different layering strategies.
I could imagine something similar here.

The commutation rules for the gates should be equally valid for all cases :-)

test/hybridmap/test_hybridmap.cpp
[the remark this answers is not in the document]
burgholzer: All the `qc.dump...` methods simply take an `std::ostream&` as an argument (note the missing `f` in the type). You can simply pass `std::cout` as an argument and that should work.

test/python/test_hybridmap.py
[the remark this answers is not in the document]
burgholzer: I think you were on the right track in that commit where you were using `__file__`. You just didn't get the path quite right I believe. IIRC, it should be `Path(__file__).parent.parent` to get out of the `python` and into the `test` folder. Maybe worth giving it one more try

test/python/test_hybridmap.py
```diff
@@ -0,0 +1,94 @@
...
+    qc = QuantumCircuit(6)
+    qc.h(0)
+    qc.cx(0, 1)
+    qc.cx(0, 2)
+    qc.cx(0, 3)
+    qc.cx(0, 4)
+    qc.cx(0, 5)
+
+    mapper.map(qc)
+    mapper.schedule(create_animation_csv=False)
```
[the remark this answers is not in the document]
burgholzer: Maybe my answer to the comment above helps to set it up properly. Then it should become a little less anoying.
But, in general, I agree that the C++ tests already cover a bit more of the functionality and it might be fine to not assert too much.
Maybe at least assert a non-zero final fidelity or something like that. Or a non-empty final circuit.

include/hybridmap/AodScheduler.hpp
lsschmid: Fixed it with const ref members. Is this how one should do this in general, or should I just always pass the architecture by argument?

include/hybridmap/Mapping.hpp
```diff
@@ -0,0 +1,122 @@
...
+public:
+  Mapping() = default;
+  Mapping(size_t nQubits, InitialMapping initialMapping) {
+    switch (initialMapping) {
+    case Identity:
+      for (size_t i = 0; i < nQubits; ++i) {
+        circToHw.insert({i, i});
+      }
+      break;
+    }
```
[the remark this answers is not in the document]
lsschmid: Yes, the idea is have it "ready" for other initial mappings.

include/hybridmap/AodScheduler.hpp
burgholzer: I'd have to see in the actual code to check (just in my phone right now), but in general one should avoid reference members in a class (i.e., `NeutralAtomArchitecture& arch`).
The main reasoning for that is written down here: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html
What you would want to do is to make the member a pointer to an architecture (i.e., `NeutralAtomArchitecture* arch`), have the constructor take a reference `NeutralAtomArchitecture& architecture`, and use `arch(&architecture)` to set the pointer member in the constructor based on the address (`&`) of the passed architecture.

include/hybridmap/NeutralAtomDefinitions.hpp
```diff
@@ -0,0 +1,40 @@
...
+using CoordIndex   = std::uint32_t;
+using CoordIndices = std::vector<CoordIndex>;
+// A HwQubit corresponds to an atom in the neutral atom architecture. It can be
+// used as qubit or not and occupies a certain position in the architecture.
+using HwQubit                      = uint32_t;
+using HwQubits                     = std::set<HwQubit>;
+using HwPositions [[maybe_unused]] = std::vector<HwQubits>;
+// A Qubit corresponds to a qubit in the quantum circuit. It can be mapped to a
+// hardware qubit.
+using Qubits = std::set<Qubit>;
```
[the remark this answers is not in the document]
lsschmid: The reason for that is that op->getUsedQubit returns a set and I didn't bother converting. Is this still worth it even if I always have to convert?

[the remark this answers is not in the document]
burgholzer: Ah, I see. In that case, not immediately.
It might be worth to add another utility function to the `Operation` class though that returns an `std::unordered_set`.

include/hybridmap/HybridNeutralAtomMapper.hpp
lsschmid: In the beginning I thought it might only be temporary, therefore the comment. I now added a docstring and included it as a regular function, as it is helpful for debugging.

include/hybridmap/HybridNeutralAtomMapper.hpp
lsschmid: Probably some left-over I guess. Removed it.

include/hybridmap/HybridNeutralAtomMapper.hpp
lsschmid: The function might also be interesting for non binding things, so I might leave it here. I renamed it so it's use becomes more clear.

include/hybridmap/HybridNeutralAtomMapper.hpp
lsschmid: I have the two stages sperate from each other. This is just a shortcut if one wants to do (as commonly done) both and for the Python bindings.

include/hybridmap/HybridNeutralAtomMapper.hpp
lsschmid: I use these functions for the python bindings, so Clion does not detect the use and therefore I added the [[maybe_unused]] part.

include/hybridmap/HybridNeutralAtomMapper.hpp
lsschmid: See below

include/hybridmap/Mapping.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+  /**
+   * @brief Returns the hardware qubits assigned to the given circuit qubits.
+   * @param qubits The circuit qubits to be queried
+   * @return The hardware qubits assigned to the given circuit qubits
+   */
+  [[nodiscard]] std::set<HwQubit> getHwQubits(std::set<Qubit>& qubits) const {
+    std::set<HwQubit> hwQubits;
+    for (const auto& qubit : qubits) {
+      hwQubits.insert(this->getHwQubit(qubit));
```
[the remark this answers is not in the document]
lsschmid: I replaced insert/push back in all files.

include/hybridmap/NeutralAtomLayer.hpp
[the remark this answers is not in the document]
lsschmid: I created an issue on the unification of the layer. But for the moment I would leave the code as it is

include/hybridmap/NeutralAtomScheduler.hpp
lsschmid: Good point. I added the animation later and didn't bother but in the end it was larger than expected. I'll try to move it in a separate file.

include/hybridmap/AodScheduler.hpp
lsschmid: Renamed it

include/hybridmap/NeutralAtomDefinitions.hpp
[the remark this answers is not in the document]
lsschmid: Ok, then I would leave it for the moment, if that's fine

test/hybridmap/test_hybridmap.cpp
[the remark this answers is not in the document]
lsschmid: Ah, didn't know that. Fixed it 👍

test/python/test_hybridmap.py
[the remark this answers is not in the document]
lsschmid: Maybe, but the way I had it, it worked on my local machine. So even if it would work with the changed path in the CI, it would probably break my local version which is also non-optimal.

test/python/test_hybridmap.py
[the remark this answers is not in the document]
lsschmid: I added some basic asserts. Might becomes irrelevant anyway with the comment above

test/CMakeLists.txt
```diff
@@ -24,3 +24,9 @@ if(TARGET MQT::LogicBlocks)
   package_add_test(mqt-logic-blocks-test MQT::LogicBlocks
                    ${CMAKE_CURRENT_SOURCE_DIR}/test_logicblocks.cpp)
 endif()
+
+if(TARGET MQT::QMapHybrid)
+  file(COPY ${PROJECT_SOURCE_DIR}/test/hybridmap DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
```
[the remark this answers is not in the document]
lsschmid: Changed that, hopefully it works.

include/hybridmap/NeutralAtomUtils.hpp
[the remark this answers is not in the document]
lsschmid: Makes sense, so ill switch to numeric_limits::max()

include/hybridmap/NeutralAtomUtils.hpp
```diff
@@ -0,0 +1,276 @@
...
+    return !(*this == other);
+  }
+
+  /**
+   * @brief Append a single move to the end of the combination.
+   * @param addMove The move to append
+   */
+  void append(AtomMove addMove) {
+    moves.push_back(addMove);
+    cost = std::numeric_limits<fp>::quiet_NaN();
```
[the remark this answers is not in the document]
lsschmid: Changed that as mentioned in another comment

include/hybridmap/Readme.md
lsschmid, line 1: I forgot about this file. This was not for documentation but as a help for Sunghye. I'll just delete it here.

src/hybridmap/HardwareQubits.cpp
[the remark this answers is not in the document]
lsschmid: This could be possible. Might be a good general idea when we talk about the unified architecture 👍

src/hybridmap/Mapping.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include "hybridmap/Mapping.hpp"
+
+#include "hybridmap/NeutralAtomDefinitions.hpp"
+
+#include <stdexcept>
+
+namespace qc {
+void Mapping::applySwap(Swap swap) {
+  auto q1 = swap.first;
+  auto q2 = swap.second;
```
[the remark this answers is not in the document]
lsschmid: Yes, I know this "trick". But for some reasons this gives me problems with Clion debugging, as it can not resolve the second version. So probably you can find multiple such things in my code, same for for loops. Not sure if this problem only exists on my end.

src/hybridmap/NeutralAtomLayer.cpp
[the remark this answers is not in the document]
lsschmid: I updated it in all files.

include/hybridmap/HybridNeutralAtomMapper.hpp
burgholzer: Fair enough. Maybe I would have envisioned more of a `NeutralAtomCompiler` meta class that orchestrates Synthesis, Mapping, Scheduling, Conversion, etc. instead of having this "mangled" in the Neutral Atom Mapper. I guess it's not really necessary to change anything here right now, but It could be helpful in the future to have some top level orchestration class that manages everything.

Just as context: we have something like that in QCEC, where there is an `EquivalenceCheckingManager` that orchestrates the whole flow and the individual and isolated `EquivalenceChecker`s that are being managed. Makes for a very clean separation of concerns.

include/hybridmap/HybridNeutralAtomMapper.hpp
burgholzer: I see. That kind of makes sense. Then only one small comment here: Consider using the same name for the function as in the other class. Consistency is key. Right now, this is using `getInitHwPos` and `getInitialHwPos`.

include/hybridmap/NeutralAtomDefinitions.hpp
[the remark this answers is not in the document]
burgholzer: Sure, feel free to resolve here. I doubt that this really is a performance bottleneck in the current mapper. It might just be worth keeping in mind.

test/python/test_hybridmap.py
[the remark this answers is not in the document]
burgholzer: Getting the paths right is always tricky. In the end, this kind of duplication is not the end of the world. So I'd be ok with just leaving it the way it is. However, there is definitely a way to fix this that only requires moderate effort and is expected to work locally and in CI.

src/hybridmap/HardwareQubits.cpp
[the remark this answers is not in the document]
burgholzer: Would you mind also creating an issue for that (maybe linking to this discussion here)?
So particularly on the architecture also managing its state.

src/hybridmap/Mapping.cpp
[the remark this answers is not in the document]
burgholzer: To be fair, I never really had problems with that but I could see why structured bindings (that's the name for the "trick" ;o) ) could become confusing for debuggers.
In the end it's mostly syntactic sugar and creates easier to read code (typically), but it does not affect performance. So feel free to just leave it as-is.

include/hybridmap/HybridNeutralAtomMapper.hpp
lsschmid: Makes sense

include/hybridmap/HybridNeutralAtomMapper.hpp
lsschmid: Yes, worked 👍

include/hybridmap/HybridNeutralAtomMapper.hpp
lsschmid: Added it to the parameters

include/hybridmap/HybridNeutralAtomMapper.hpp
lsschmid: Definitely would make sense. I think the reason to it this way was also because I cannot convert the mapped circuit back to Python (Qiskit) as it is no longer valid qasm. Once we manage to unify this using e.g. Opaque gates, this should be no problem.

include/hybridmap/HybridNeutralAtomMapper.hpp
lsschmid: You are right, didn't notice that.

include/hybridmap/NeutralAtomArchitecture.hpp
```diff
@@ -0,0 +1,485 @@
...
+   * by an edge in the resulting connectivity graph. This can be be computed
+   * beforehand.
+   */
+  void computeNearbyCoordinates();
+
+public:
+  std::string name;
+
+  // Constructors
+  NeutralAtomArchitecture() = delete;
```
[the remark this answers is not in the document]
lsschmid: As an architecture without the json doesn't really make sense I thought I might prevent it. But can also allow it obviously.

include/hybridmap/NeutralAtomArchitecture.hpp
```diff
@@ -0,0 +1,485 @@
...
+            this->coordinates[idx2].getXfp(), this->coordinates[idx2].getYfp()};
+  }
+  /**
+   * @brief Computes the time it takes to move a qubit along a MoveVector
+   * @param v The MoveVector
+   * @return The time it takes to move a qubit along the MoveVector
+   */
+  [[nodiscard]] fp getVectorShuttlingTime(const MoveVector& v) const {
+    return v.getLength() * this->getInterQubitDistance() /
+           this->getShuttlingTime(OpType::Move);
```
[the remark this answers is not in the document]
lsschmid: I kinda prefer length compared to size for a 2D vector

src/hybridmap/AodScheduler.cpp
lsschmid: Is is due to the tests. This happens very occasionally, but we had it during the evaluations of the DAC paper. Unfortunately, I don't remember the settings, and these situations are very hard to create manually.

src/hybridmap/HybridNeutralAtomMapper.cpp
lsschmid: The check prevent commuting gates to be applied multiple times. The TODO was because I thought it also might work without it, but as a result I think this is the best solution. Therefore I just removed the TODO

src/hybridmap/HybridNeutralAtomMapper.cpp
lsschmid: It actually might happen but these are edge cases which are really hard to construct by hand, so for the moment I would keep it as it is. Once I find a good test, I'll add it.

lsschmid: I'll try to update the test such this edge case may occure

src/hybridmap/HybridNeutralAtomMapper.cpp
lsschmid: Same here

src/hybridmap/HardwareQubits.cpp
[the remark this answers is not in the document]
lsschmid: Done 👍 
I also added a reset method to the mapper which basically does more or less that.

src/hybridmap/Mapping.cpp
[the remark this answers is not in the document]
lsschmid: Ok, then I'll leave as it is, although I like the version with structured bindings more, but it can be really annoying during debugging.

test/CMakeLists.txt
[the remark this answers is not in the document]
lsschmid: It did not work, although I'm not sure why. On my local machine the new version works fine but the CI complains it can't find the files so I reverted to the original version. If you know what the problem is, let me know.

test/python/test_hybridmap.py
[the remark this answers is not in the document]
lsschmid: There should be a solution I guess. But debugging the CI is almost impossible (or at least I don't know how).

include/hybridmap/Readme.md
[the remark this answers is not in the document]
lsschmid: Actually forgot about this file. This was not for official documentation but meant as a help for Sunghye. Removed it now.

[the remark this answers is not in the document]
lsschmid: I also created an Issue here #466

include/hybridmap/NeutralAtomArchitecture.hpp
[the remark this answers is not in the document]
burgholzer: Given how C++ sometimes works, it's probably better to make the architecture default constructible. I see practically no downside and a slight plus in the flexibility. In the future there might be even more constructors that do not require a file/filesystem access at all (but, e.g., use QDMI to populate the individual properties). There it might be nice to first default construct and then populate.

include/hybridmap/NeutralAtomArchitecture.hpp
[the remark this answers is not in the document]
burgholzer: Fine with me. Just wanted to bring it up 👍🏼 feel free to resolve here

src/hybridmap/AodScheduler.cpp
burgholzer: I see. Then, there is no real action item here for now. Would be nice to maybe extend the tests in the future to cover more scenarios. Could just include a couple more circuits 🤷🏼

test/CMakeLists.txt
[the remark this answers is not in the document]
burgholzer: I'll see whether I can resolve that locally. It's probably about the working directory when running the tests. These things are really hard to get right and always make problems. That's why I much prefer circuits to be defined in the C++ or Python sources themselves which completely eliminates these problems.
Same goes for the architectures. Here one could create an `std::stringstream` with the config to avoid the filesystem...
I'll try to have a look once most of the other conversations have been resolved.

test/python/test_hybridmap.py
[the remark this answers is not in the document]
burgholzer: I'll add this to the list of things that I'll try to check once most other things are resolved. Then I can cover the C++ tests and files together with the Python tests. I've done my fair share of fiddling around with CI and the tests in the past. So, ideally, this will be a rather quick fix.

src/hybridmap/Mapping.cpp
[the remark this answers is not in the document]
burgholzer: Maybe worth to investigate your debugging setup at some point. It should definitely be possible to get that working.
Quite naturally, debugging should happen on a `Debug` build. Maybe it's one of the compiler flags we enable that prevents a good debugging experience. Something to look into. Debuggability should not be the limiting factor for writing nice code.
Feel free to resolve this for now.

include/hybridmap/HybridNeutralAtomMapper.hpp
burgholzer: Yeah, I see where you are coming from.
Let's keep this in mind when we discuss the general unification for the compiler collection. Feel free to resolve for now.

lsschmid:
The CI fails on Windows with some weird error. 
When I install and run the tests on a Windows VM, everything works fine. Not sure what to do?!

include/hybridmap/NeutralAtomArchitecture.hpp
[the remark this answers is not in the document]
lsschmid: Added back the default constructor 👍

include/hybridmap/NeutralAtomUtils.hpp
```diff
@@ -0,0 +1,276 @@
...
+
+  [[nodiscard]] fp operator()(uint32_t row, uint32_t col) const {
+    if (row < col) {
+      return data[col][row];
+    }
+    return data[row][col];
+  }
+
+  [[nodiscard]] uint32_t getSize() const { return size; }
+};
```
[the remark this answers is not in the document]
lsschmid: Moved it to mqt-core and used a template 👍

src/hybridmap/AodScheduler.cpp
lsschmid: I included only a few to keep the overhead as small as possible. In general one can of course add more.
One could also try around with the random seed, probably there is one which creates some of these special cases.

include/hybridmap/NeutralAtomDefinitions.hpp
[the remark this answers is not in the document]
lsschmid: I'll remember it for the future. When I write the code I barely knew anything about C++, so there a lot of things I would do differently next time 👍

include/hybridmap/NeutralAtomScheduler.hpp
lsschmid: Done, now the creation of the animation is in a Animations.hpp

src/hybridmap/AodScheduler.cpp
burgholzer: Maybe consider adding one more test circuit and/or architecture for now 🤷🏽‍♂️ it's fine if CI takes a couple seconds longer.

include/hybridmap/AodScheduler.hpp
lsschmid: For cases where I don't change the arch and I don't want to copy the object I kept the ref members, in the other cases I used a pointer.

burgholzer (approved):
Alright, all conversations have been resolved, the CI is green, and follow-up issues have been created. ✅ 
So I think we are good to go here for now! 🎉 


Thanks for all your work on the PR! 🙏🏼 
I'll merge once CI passes.

PR #469, opened 2024-06-05
💚 Resolve irregularly failing tests

ystade:
## Description

The test of the NA-mapper failed from time to time. The cause seems to be the size of the entangling zone that is too narrow when the mapper comes up with a suboptimal run.

## Checklist:


ystade:
@burgholzer I substituted all vector accesses that happened with the []-operator with a call to the method `.at()` to enable bound checks, see https://github.com/cda-tum/mqt-qmap/pull/469/commits/0f40f707a14b125daebb5c671793f937fc52d8c2. Not sure whether we want to keep this because of performance considerations.

The issue of the failing test was actually not a Windows problem; on other machines, the error could have happened as well and would have resulted in a segmentation fault. I adapted the test case now accordingly.

burgholzer (approved):
Thanks for digging into this!
Overall looks good to me. 
I just skimmed the changes and added suggestions where I deemed that the range checked access is definitely not necessary because a range check is already conducted before that.
You should be able to simply batch up all suggestions into a single commit on GitHub.
I'd keep the other at's just for safety.

ystade:
> Thanks for digging into this! Overall looks good to me. I just skimmed the changes and added suggestions where I deemed that the range checked access is definitely not necessary because a range check is already conducted before that. You should be able to simply batch up all suggestions into a single commit on GitHub. I'd keep the other at's just for safety.

Thanks a lot for going through them thoroughly. I will add these suggestions.

ystade:
I am working on proper tests right now that actually test the functionality (not only the termination). This I think makes more sense after all.

burgholzer:
> I am working on proper tests right now that actually test the functionality (not only the termination). This I think makes more sense after all.

Makes sense. I'd guess you'd prefer to do that in a separate PR, right? Or should I wait before merging this one?

ystade:
> > I am working on proper tests right now that actually test the functionality (not only the termination). This I think makes more sense after all.
> 
> Makes sense. I'd guess you'd prefer to do that in a separate PR, right? Or should I wait before merging this one?

I'll open a new PR for that soon.

PR #473, opened 2024-06-14
Python bindings, keeping arch alive

lsschmid:
## Description

This fixed the problem that for the Hybridmapper, using the Python bindings it can happen that the architecture object is deleted by the garbage collector, loosing the reference for the mapper. Now it is kept alive using the keep_alive functionality of PyBind.

## Checklist:


extern/mqt-core
burgholzer, line 1: I think this change might be unintentional.

burgholzer (approved):
Looks like an easy fix. Great 👍🏻

I think the submodule change is unintentional though, so consider reverting that part of the PR. 😌

PR #474, opened 2024-06-14
Keep na arch alive 2

lsschmid:
## Description

This fixed the problem that for the Hybridmapper, using the Python bindings it can happen that the architecture object is deleted by the garbage collector, loosing the reference for the mapper. Now it is kept alive using the keep_alive functionality of PyBind.

## Checklist:


lsschmid:
I'm not sure. Clion tells me that mqt-core has a newer version (which is true). But when I commit the exter link to mqt-core, it modifies the .pre_commit. Not sure why, but for this PR it also doesn't really matter.

burgholzer:
> I'm not sure. Clion tells me that mqt-core has a newer version (which is true). But when I commit the exter link to mqt-core, it modifies the .pre_commit. Not sure why, but for this PR it also doesn't really matter.

The best thing to do in cases like these is to simply not commit the submodule change.
If you want to make sure that you are running the same submodule version as on main, you can do
```
git submodule update --init --recursive 
```

PR #496, opened 2024-07-31
Fix Qubit out of Bounds Error

ystade:
## Description

When executing slightly larger circuits it could happen that the size, particularly the width, of the entangling did not suffice which caused and uncaught error. This PR resolves this problem by disregarding all qubits that do not fit in the entangling zone and leaving them for a later iteration.

## Checklist:


ystade:
@burgholzer You can already look over the changes. I will fix the Linter warnings tomorrow.

ystade:
@burgholzer I was apparently to quick, i will first fix the failing tests... and ping you then.

cmake/ExternalDependencies.cmake
```diff
@@ -29,7 +29,7 @@ endif()
 # cmake-format: off
 set(MQT_CORE_VERSION 2.5.1
     CACHE STRING "MQT Core version")
-set(MQT_CORE_REV "35e06ca3067ca3cf36bda1f0c38edf5bd7456fb6"
+set(MQT_CORE_REV "633915adeece6de3001dd16563bea9d321112b10"
```
burgholzer, line 32:
```suggestion
set(MQT_CORE_REV "aa3b3bd3b07e43d6421b7c0d741b5755938c6e9d"
```
Should be the up-to-date rev

cmake/ExternalDependencies.cmake
```diff
@@ -29,7 +29,7 @@ endif()
 # cmake-format: off
 set(MQT_CORE_VERSION 2.5.1
```
burgholzer, line 30:
```suggestion
set(MQT_CORE_VERSION 2.5.2
```
Given that this is no longer pointing to the commit that is tagged as `2.5.1`, this should be bumped as well. Otherwise, the next update run will compare the versions and overwrite the rev with the `2.5.1` release one.

include/na/NAGraphAlgorithms.hpp
```diff
@@ -152,7 +152,8 @@ class NAGraphAlgorithms {
    * @return A tuple of vectors containing the moveable vertices, the fixed
    * vertices mapped to their relative x-position in every time step.
    */
-  [[nodiscard]] static auto computeSequence(const InteractionGraph& g)
+  [[nodiscard]] static auto computeSequence(const InteractionGraph& g,
+                                            std::size_t             maxSites)
```
burgholzer, line 156: The docstring of this method should be updated to reflect the additional parameter.

src/na/Architecture.cpp
```diff
@@ -545,7 +546,14 @@ auto Architecture::getPositionOffsetBy(const Point& p, const Number& rows,
     if (!nextSiteOpt.has_value()) {
       break;
     }
-    anchorSitePos = getPositionOfSite(nextSiteOpt.value());
+    const auto& newAnchorSitePos = getPositionOfSite(nextSiteOpt.value());
+    // The following check is used to decide whether the patch is located
+    // outside of the zone Patches that overlap the border of the zone are not
+    // possible with this approach yet
+    if (std::abs(newAnchorSitePos.y - anchorSitePos.y) < std::abs(dy)) {
+      break;
```
burgholzer, line 554: Could it be that this line is uncovered by the existing tests?
Would be great if additional checks like that would be covered by tests so that we know they are effective.

src/na/Architecture.cpp
```diff
@@ -555,7 +563,11 @@ auto Architecture::getPositionOffsetBy(const Point& p, const Number& rows,
     if (!nextSiteOpt.has_value()) {
       break;
     }
-    anchorSitePos = getPositionOfSite(nextSiteOpt.value());
+    const auto& newAnchorSitePos = getPositionOfSite(nextSiteOpt.value());
+    if (std::abs(newAnchorSitePos.x - anchorSitePos.x) < std::abs(dx)) {
+      break;
+    }
```
burgholzer, line 569: Would it make sense to also add a similar code comment like above here?

src/na/NAGraphAlgorithms.cpp
```diff
@@ -188,11 +188,42 @@ auto NAGraphAlgorithms::colorEdges(
...
                 const auto w = b.first == v ? b.second : b.first;
-                return partialOrder.isReachable(u, w) or
-                       (!partialOrder.isReachable(w, u) and
-                        (nAdjColors[a] > nAdjColors[b] or
-                         (nAdjColors[a] == nAdjColors[b] and
-                          edgeDegree[a] > edgeDegree[b])));
+                if (u == w) {
+                  // the compare function defines a proper less than relation,
+                  // i.e., equal elements must return false
+                  return false;
```
burgholzer, line 194: This is not yet covered by tests as well. Would that be hard to do?

src/na/NAGraphAlgorithms.cpp
```diff
@@ -437,11 +469,14 @@ auto NAGraphAlgorithms::computeSequence(const InteractionGraph& g)
...
             });
-  const auto& sequence = groupByConnectedComponent(g, sequenceUngrouped);
-  const auto& [coloring, partialOrder] =
+  auto        sequence = groupByConnectedComponent(g, sequenceUngrouped);
+  const auto& colorEdgesResult =
       colorEdges(g, coveredEdges(g, maxIndepSet), sequence);
-  const auto& fixed   = partialOrder.orderTopologically();
-  const auto& resting = computeRestingPositions(sequence, fixed, coloring);
+  std::unordered_map<Edge, Color, qc::PairHash<qc::Qubit, qc::Qubit>> coloring =
+      colorEdgesResult.first;
```
burgholzer, line 476:
```suggestion
  auto coloring = colorEdgesResult.first;
```
Just a suggestion because the type is so verbose

burgholzer (changes requested):
Overall this looks really good.
I just have a few minor comments, which you will find inline.
I hope they are easy to address.

src/na/NAGraphAlgorithms.cpp
ystade: So, this is only written this way with all the if-else branches because clang-tidy was warning because of a long condition. I tried to take benefit of the new structure and added helpful comments for each case. Additionally, testing it, is hard because I would need to engeneer for each case a special circuit and then it also depends on the randomness of memory addresses. I am very hesitating to do that now, in particular as the logica of the code did not change really.

burgholzer: I really like the new structure of this.
Makes things quite explicit 👍🏻

And just to make that clear: it is only this single if statement that is not yet covered. All the others are covered already according to the data.

However, I fully understand that you do not want to spend a substantial amount of time to engineer an example.
If not quickly possible, then feel free to just resolve this conversation 😌

burgholzer (approved):
Looking great 👍🏼

Issue #328, opened 2023-06-02
✨ Connectivity constraints for the Clifford synthesis


pehamTom:
### What's the problem this feature will solve?

The current encoding for the CNOT-constraints in the Clifford synthesizer assumes all-to-all connectivity. Many quantum computing platforms only offer limited qubit connectivity.

### Describe the solution you'd like

Allow passing a coupling map to the synthesizer. A class for this already exists within QMAP. The only thing that changes is the encoding loop for the two-qubit constraints. 

It would be of general interest to see how a constant limited connectivity might help scalability to larger circuits as the CNOT constraints are the (asymptotically) dominant factor for the size of the encoding.

burgholzer:
Closing as not planned for now.

PR #506, opened 2024-08-10
Connectivity constraints for clifford synthesis

JakobSchaeffeler:
## Description
This PR implements connectivity constraints for the clifford synthesis as requested by #328, by passing optional coupling maps. The passed coupling map used to restrict CNOTs to only connected qubits. Additionally, the clifford synthesizer needs to check if the circuit can be synthesized with any permutation of qubits. If no coupling map is passed all to all connectivity is assumed. 

## Checklist:


burgholzer:
> [!NOTE]
> #507 just went in and brought some updates from the MQT Core dependency that cause some conflicts here. Could you please resolve these? They should be fairly easy to resolve. Sorry for the inconvenience.

JakobSchaeffeler:
@burgholzer I tried to resolve the conflicts and everything works locally. However the CI/Change/Check test fails, but all subtasks in there are marked as successful. Do you know what the issue is?

burgholzer:
> @burgholzer I tried to resolve the conflicts and everything works locally. However the CI/Change/Check test fails, but all subtasks in there are marked as successful. Do you know what the issue is?

Thanks for fixing the conflicts so quickly.
I just retriggered the respective CI job. Let's see if it still fails on the second try. 
Should be unrelated to your changes in any case.

include/cliffordsynthesis/Results.hpp
```diff
@@ -103,6 +130,8 @@ class Results {
   double runtime = 0.0;
   std::size_t solverCalls = 0U;
 
+  std::string pvals;
+  std::vector<std::vector<bool>> pvector;
```
pehamTom, line 134: I am assuming the p here stands for permutation? Maybe just write this one out, so it is immediately clear.

src/cliffordsynthesis/Tableau.cpp
```diff
@@ -492,4 +492,107 @@ bool Tableau::isIdentityTableau() const {
   }
   return true;
 }
+
+Tableau Tableau::applyMapping(const std::vector<std::vector<bool>>* p) {
```
pehamTom, line 496: No features of a pointer are used in this function. In this case it is better to just use a `const` reference.

src/cliffordsynthesis/Tableau.cpp
```diff
@@ -492,4 +492,107 @@ bool Tableau::isIdentityTableau() const {
...
+  Tableau mappedTableau = Tableau(nQubits, hasDestabilizers());
+  for (size_t i = 0; i < mappedTableau.getTableauSize(); i++) {
+    for (unsigned char& j : mappedTableau.tableau[i]) {
+      j = 0;
+    }
+  }
```
pehamTom, line 502: Is the zeroing out necessary? In the next loops, don't you overwrite every entry of the mapped tableau anyway?

src/cliffordsynthesis/Tableau.cpp
```diff
@@ -492,4 +492,107 @@ bool Tableau::isIdentityTableau() const {
...
+          }
+        }
+      }
+    }
+    pivot++;
+  }
+}
+// equivalence check for two Tableaus without destablilizers
+bool Tableau::equivalentUpToStabilizer(const Tableau* t) const {
+  assert(getTableauSize() == t->getTableauSize());
```
pehamTom, line 587: In this case, we could also just return `false` if the sizes differ

src/cliffordsynthesis/Tableau.cpp
```diff
@@ -492,4 +492,107 @@ bool Tableau::isIdentityTableau() const {
...
+        mappedTableau.tableau[i] = tableau[j];
+        mappedTableau.tableau[i + nq] = tableau[j + nq];
+      }
+    }
+  }
+  return mappedTableau;
+}
+
+// in place Gauss Elimination
+void Tableau::gaussianEliminationGF2() {
```
pehamTom, line 542: Minor comment: since you compute the reduced row echelon form of the matrix and use it's uniqueness to compare tableaus, calling this method something like `rref` would make it clearer what is actually computed.

src/cliffordsynthesis/encoding/GateEncoder.cpp
```diff
@@ -135,12 +136,12 @@ void GateEncoder::assertZConstraints(const std::size_t pos,
                                      const std::size_t qubit) {
   const auto& gatesToZTransformations = [this](const auto& p1, const auto& p2,
                                                const auto& p3) {
-    return tvars->singleQubitZChange(p1, p2, p3);
+    return tvars->singleQubitZChange(p1 + 1, p2, p3);
```
pehamTom, line 139: Very briefly, for my understanding, these offsets are due to the additional "permutation tableau", right?

src/cliffordsynthesis/encoding/TableauEncoder.cpp
```diff
@@ -68,6 +82,48 @@ void TableauEncoder::assertTableau(const Tableau& tableau,
...
+      sumRow = sumRow + vars.p[i][j];
+    }
+    lb->assertFormula(sumRow == LogicTerm(vr1));
+    const int32_t vc = 0;
+    const int32_t vc1 = 1;
+    LogicTerm sumCol = LogicTerm(vc);
+    for (std::size_t j = 0U; j < N; ++j) {
+      sumCol = sumCol + vars.p[j][i];
+    }
+    lb->assertFormula(sumCol == LogicTerm(vc1));
```
pehamTom, line 112: Instead of doing symbolic arithmetics here, consider using the `exactlyOneCmdr` function (also used in the `GateEncoder` class.

pehamTom (commented):
Very nice work! 

I actually do not have too much to comment on this. The only thing that could be nicer from a maintainability standpoint is to get rid of the offsets in the encodings of every constraint. That might require some more involved rewrites. Unless you have a quick and elegant idea to get rid of this, we can leave it for future refactoring.

One additional feature that would go a long way is to also include a small example for this functionality in the documentation. Extending the Jupyter Notebook for https://mqt.readthedocs.io/projects/qmap/en/latest/Synthesis.html by a small example should be sufficient.

src/cliffordsynthesis/encoding/GateEncoder.cpp
JakobSchaeffeler: That's right, I also tried implementing it without offsets and then forcing that no Gates are applied in the first and last timestep. But then unnecessary variables for gates are created and need to be set to False, so I chose to use offsets way.

JakobSchaeffeler:
@burgholzer @pehamTom I incorporated the feedback by Tom. However now all Python Test fail with a Deprecation warning and one C test which should be unrelated to my PR. Do you know how to solve that?

src/cliffordsynthesis/encoding/GateEncoder.cpp
pehamTom: Gotcha, given that there is some other encoding techniques that need to be worked into QMAP we just leave that for a future refactoring

pehamTom:
> @burgholzer @pehamTom I incorporated the feedback by Tom. However now all Python Test fail with a Deprecation warning and one C test which should be unrelated to my PR. Do you know how to solve that?

Yeah, that is not your fault. We leave the PR open until the IDP is over anyway. I will take care of it.

burgholzer:
> @burgholzer @pehamTom I incorporated the feedback by Tom. However now all Python Test fail with a Deprecation warning and one C test which should be unrelated to my PR. Do you know how to solve that?

I think if you hit the "Update Branch" button (effectively merging in the main branch), the deprecation warnings should be gone. IIRC, I fixed them for QMAP in one of the latest updates.
Otherwise, I'll probably address them next week in a separate PR. They should not be related to your changes here 😌

burgholzer:
Just merged the main branch in which contains the fixes from #509.
The deprecation warnings here are now really related to your code changes and should be addressed (should be fairly easy).

burgholzer:
@pehamTom
> I will take care of it.

Is this still true? Would be nice to not see this go to waste.

pehamTom:
Yes this has gone a bit under the radar. Given that most developments regarding Clifford circuit synthesis are taking place in QECC now, I would move this there. I am actually working on very similar things for state preparation/encoding circuit synthesis and have this implemented for CSS codes (pure CNOT circuits) on a branch that is not ready to be merged yet. To avoid duplication of functionality and given that this would also be a lot of refactoring work in qmap I would close this PR.

burgholzer:
> Yes this has gone a bit under the radar. Given that most developments regarding Clifford circuit synthesis are taking place in QECC now, I would move this there. I am actually working on very similar things for state preparation/encoding circuit synthesis and have this implemented for CSS codes (pure CNOT circuits) on a branch that is not ready to be merged yet. To avoid duplication of functionality and given that this would also be a lot of refactoring work in qmap I would close this PR.

I see. Thanks for the context.
Given the different demands for FTQC and near-term quantum computing (for example, in terms of the gates that can be used), I could still see value in keeping the Clifford synthesis that is already in QMAP and, potentially, even extending it.
Quite a bit of work has already been put into this, and I fell it would be a shame to not see this through.
Could you maybe summarize what you think is still missing here? (besides an obvious merge from main to resolve the conflicts)
It does not have to be you who does the actual work then.

pehamTom:
The circuit synthesis in QECC is not entirely tailored towards FTQC. Some methods make specific circuits fault-tolerant, but these things are kept separate. Regarding gates, the restriction is also not there. The circuits in QECC are on the physical level, not the logical level. Here in QMAP we also do not support arbitrary single- or two-qubit Clifford gates. 

I agree that seeing the work on Clifford circuit synthesis in QMAP vanish is not desirable, and I would certainly keep the current implementation. On the other hand, some things are very cumbersome and unnecessary. The maxsat encoding, for example, is, in practice, strictly worse than the iterative procedure. Furthermore, the synthesis for states (tableaus without destabilizers) is not ideal since it doesn't guarantee actually compiling optimal circuits (they are only optimal for the given tableau). The compilation in QECC has an improved encoding for these things. For future developments, I would, therefore, prefer to keep developments in a single repository. 

That being said, the functionality implemented in this PR is pretty much "hard coded" into the existing implementation and would need to be separated. For connectivity-aware encoding, two extra tableaus are needed to encode the effect of the initial state mapping. In principle, one would need to refactor the `TableauEncoder` as well as the other encoder classes. One could also avoid this by assuming a graph with all-to-all connectivity by default and manually checking whether this has been changed. If yes, one can leave out the extra steps required for encoding the mapping.

burgholzer:
Based on some internal discussions, I am closing this for now.
A lot of development on Clifford synthesis is currently going on over in https://github.com/cda-tum/mqt-qecc and it would be wasteful to duplicate the effort here.
At some point, the corresponding functionality shall be unified again.

PR #556, opened 2024-11-15
🚸 Python Bindings for NA State Preparation

ystade:
## Description

This PR aims to provide python bindings for the state preparation for neutral atoms part of qmap.

## Checklist:


ystade:
@burgholzer The state of that PR is far from being ready. However, I seem to make a major mistake since when I first execute `uv pip install --editable ".[dev]"` and then run `from mqt.qmap import na` in that environment, this causes the error "ModuleNotFoundError: No module named 'mqt.qmap.na.state_preparation'".

Can you provide a hint what I am doing wrong here?

burgholzer:
> Can you provide a hint what I am doing wrong here?

Quickly fixed that for you in the code in https://github.com/cda-tum/mqt-qmap/pull/556/commits/5e600c325b2b684d3362644df0c1b06129fe8e20
When you bind C++ code in the `bindings.cpp` file to the module `m`, the resulting definitions become part of the `mqt.qmap.pyqmap` module.
That is also the module that needs the stub files for proper typing support.
In the Python package itself, you then simply re-export the corresponding symbols from the compiled module.

In the future, I'd like to see the QMAP bindings reworked a little bit to make them a little more modular, taking a bit of inspiration from MQT Core and how it provides its bindings.
Given how the functionality for state prep is really just re-exported, one could technically avoid that all together by directly building a `state_preparation` compiled extension and directly installing it into the `na` directory.

Hope you'll be able to take it from here. Let me know if there is more you'd want to know!

ystade:
@burgholzer Thanks for your help. Now, I got further in the python bindings. I have now a problem that the test with the circuit `steane.qasm` passes while the same test with `surface_3.qasm` does not pass (same for `shor.qasm`) because of an `bad_alloc` or just `vector` excpetion. Do you have an idea, what the problem could be? Is it the case that the bynding to `std::vector<std::pair<qc::Qubit, qc::Qubit>>` does not work in reliable?

src/python/bindings.cpp
```diff
@@ -902,4 +909,180 @@ PYBIND11_MODULE(pyqmap, m, py::mod_gil_not_used()) {
...
+      },
+      "circ"_a, "result"_a, "max_h_offset"_a, "max_v_offset"_a,
+      "min_entangling_y"_a, "max_entangling_y"_a, "min_atom_dist"_a = 1,
+      "no_interaction_radius"_a = 10, "zone_dist"_a = 24, R"(
+Generate code for the given circuit using the solver's result. Some parameters
+of the abstraction from the 2D grid used for the solver must be provided again.
+
+.. warning::
+    The result is only consistent, when the same parameters corresponding to
+    them used for the solver.
```
burgholzer, line 1037: Just stumbled upon this while trying to understand where things could be going wrong with the bindings:
Can't you define that error out of existence by adapting this function to take the solver object as a parameter instead of just the result?

test/python/na/test_state_preparation.py
```diff
@@ -0,0 +1,57 @@
...
+    "circuit_filename",
+    [
+        "steane.qasm",
+        "surface_3.qasm",
+    ],
+)
+def test_na_state_prep_sat(solver: NAStatePreparationSolver, circuit_filename: str) -> None:
+    """Test the state preparation for the zoned neutral atom architecture."""
+    qc = QuantumCircuit.from_qasm_file(circ_dir / circuit_filename)
+    ops = get_ops_for_solver(qc, "z", 1)
```
burgholzer, line 32: Similar to the other comment: Should a user really have to call this?
Can't the solver handle this transparently when given a circuit as an input?

burgholzer:
> @burgholzer Thanks for your help. Now, I got further in the python bindings. I have now a problem that the test with the circuit `steane.qasm` passes while the same test with `surface_3.qasm` does not pass (same for `shor.qasm`) because of an `bad_alloc` or just `vector` excpetion. Do you have an idea, what the problem could be? Is it the case that the bynding to `std::vector<std::pair<qc::Qubit, qc::Qubit>>` does not work in reliable?

Oh boy.. that took a while. The failure has absolutely nothing to do with the bindings.
It's just that the `surface_3.qasm` file requires 9 qubits and you only configure the solver object with 7. Thus, you are accessing some illegal memory addresses in the `getCircuitExecutionConstraints` method.
There are a couple of things to take from this in my opinion:
- the current interface is error prone
- the code seems to miss some sanity checks that the parameters being passed to the solver make sense

src/python/bindings.cpp
ystade: Even that could go wrong, when the user provides another solver than the one that was to retrieve the result. I decided to respective information to the `Result` class.

test/python/na/test_state_preparation.py
ystade: In contrast to the other comment, I think here, this interface makes the capabilities of the solver clear. Especially, that it can not handle arbitrary circuits and that the entangling gates are taken as a list of qubit pairs. I would be in favour of leaving the interface as is.

burgholzer: I can see that point.
It still feels a bit cumbersome to use to be honest.
But I am not willing to die on that hill, so let's just keep it like it is for now 😌

docs/DevelopmentGuide.rst
```diff
@@ -179,27 +179,23 @@ Furthermore, they provide a command to automatically format your code according
 Working on the Python module
 ############################
 
-`Pybind11 <https://pybind11.readthedocs.io/>`_ is used for providing bindings of the C++ core library to Python.
+`Pybind11 <https://pybind11.readthedocs.io/en/stable/>`_ is used for providing bindings of the C++ core library to Python.
 This allows to keep the performance critical parts of the code in C++ while providing a convenient interface for Python users.
-All of the bindings code as well as the Python module itself is contained in the :code:`mqt/qmap` directory.
+All of the bindings code is contained in the :code:`src/python` directory.
+The Python package itself lives in the :code:`src/mqt/core` directory.
```
burgholzer, line 185:
```suggestion
The Python package itself lives in the :code:`src/mqt/qmap` directory.
```

src/mqt/qmap/pyqmap.pyi
```diff
@@ -631,3 +631,46 @@ class NeutralAtomHybridArchitecture:
     def nqubits(self) -> int: ...
     @property
     def nrows(self) -> int: ...
+
+class NAStatePreparationSolver:
+    def init(
```
burgholzer, line 636:
```suggestion
    def __init__(
```
Pretty sure this is also what trips up mypy.

burgholzer (approved):
Great work, especially on all the documentation! Many thanks for that!

I have nothing more to complain about 😉
The inline comments should hopefully fix the mypy warning and a copy and paste error.

PR #559, opened 2024-11-16
🐛 Fix test suite regression in heuristic mapper

burgholzer:
## Description

Due to some unknown reason, the test suite suddenly started to fail on linux systems for one particular heuristic mapper test.
The underlying issue seems to have been some concurrency issue in combination with the logging functionality in the heuristic mapper.
The fix for now was to add a random suffix to the data logger path for all respective tests.

## Checklist:


burgholzer:
This was really getting frustrating.
And it was not reproducible on macOS at all.
Due to lack of time, this PR simply drops the one problematic test.
The underlying cause is related to concurrency and how the tests are being run. 
However, I it has proven to be quite hard to really find out what is going on.
Assuming CI passes, this should be good enough for now.

burgholzer:
Alright, after some more trial and error it turns out that the error can also be fixed by adding a random suffix to the data logger path, which seems to avoid the concurrency problems.
All tests are green again now ✅

PR #665, opened 2025-06-05
🎨🐍 Streamline Python bindings

denialhaag:
## Description

This PR further streamlines the implementation of Python bindings. It is a continuation of #624.

## Checklist:

- [x] ~~I have added appropriate tests that cover the new/changed functionality.~~
- [x] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes or removals.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ QMAP is available via [PyPI](https://pypi.org/project/mqt.qmap/) for Linux, macO
 Compiling a given quantum circuit to a certain device is as easy as
 
 ```python3
-from mqt import qmap
+from mqt.qmap.plugins.qiskit.sc import compile
 from qiskit import QuantumCircuit
 from qiskit.providers.fake_provider import GenericBackendV2
 
@@ -59,13 +59,13 @@ arch = GenericBackendV2(
     num_qubits=5,
     coupling_map=[[0, 1], [1, 0], [1, 2], [2, 1], [1, 3], [3, 1], [3, 4], [4, 3]],
 )
-circ_mapped, results = qmap.compile(circ, arch=arch)
+circ_mapped, results = compile(circ, arch=arch)
 ```
 
 Optimizing a Clifford circuit is as easy as
 
 ```python3
-from mqt import qmap
+from mqt.qmap.plugins.qiskit.clifford_synthesis import optimize_clifford
 from qiskit import QuantumCircuit
 
 circ = QuantumCircuit(2)
@@ -74,7 +74,7 @@ circ.cx(0, 1)
 circ.h(0)
 circ.h(1)
 
-circ_opt, results = qmap.optimize_clifford(circ)
+circ_opt, results = optimize_clifford(circ)
 ```
 
 **Detailed documentation on all available methods, options, and input formats is available at [ReadTheDocs](https://mqt.readthedocs.io/projects/qmap).**


python/mqt/qmap/subarchitectures.py
```diff
@@ -6,33 +6,15 @@
 #
 # Licensed under the MIT License
 
-"""Functionality for computing good subarchitectures for quantum circuit mapping.
-
-This file implements the methods presented in https://arxiv.org/abs/2210.09321.
-"""
+"""Subarchitectures module for backwards compatibility."""
```
denialhaag: I needed to create this backwards-compatibility module for `ibm_guadalupe_16.pickle` and `rigetti_16.pickle` to still work. It's probably possible to recreate those files, but I didn't understand how. 😬

burgholzer: That should be fairly straightforward (although badly documented).
You can just follow the unit tests over at https://github.com/munich-quantum-toolkit/qmap/blob/main/test/python/test_subarchitectures.py#L201

Just modify that unit test accordingly to produce both of the subaechitecture pickles.

denialhaag: That worked perfectly. Thanks!

.github/workflows/ci.yml
```diff
@@ -137,11 +137,12 @@ jobs:
     if: fromJSON(needs.change-detection.outputs.run-cpp-linter)
     uses: munich-quantum-toolkit/workflows/.github/workflows/reusable-cpp-linter.yml@v1.10
     with:
-      setup-z3: true
       clang-version: 20
-      setup-python: true
+      cmake-args: -DBUILD_MQT_QMAP_BINDINGS=ON
+      files-changed-only: false
```
denialhaag:
```suggestion
      files-changed-only: true
```

denialhaag:
If I'm not mistaken, the CI is failing because we also need to pin the QCEC version once https://github.com/munich-quantum-toolkit/qcec/pull/618 is merged. 👍

python/mqt/qmap/plugins/qiskit.py
```diff
@@ -16,7 +16,7 @@
     from qiskit.providers import BackendV2
     from qiskit.transpiler import Target
 
-from ..pyqmap import Architecture
+from mqt.qmap.sc.sc import Architecture
```
burgholzer: Hm. The duplication between the bindings module and the Python module is kind of bugging me.
Something about that feels wrong to me.
What's the obstacle for directly placing these under `mqt.qmap.<...>`?

python/mqt/qmap/sc/compile.py
```diff
@@ -6,31 +6,18 @@
...
 
 from mqt.core import load
 from mqt.core.plugins.qiskit import mqt_to_qiskit
 
-from .load_architecture import load_architecture
-from .load_calibration import load_calibration
-from .pyqmap import (
+from mqt.qmap.sc.load_architecture import load_architecture
+from mqt.qmap.sc.load_calibration import load_calibration
+from mqt.qmap.sc.sc import (
```
burgholzer: I would prefer to keep the relative imports everywhere just as we do that in the other MQT packages.

burgholzer (changes requested):
This generally looks pretty much ready. I only have two points for discussion, inline.

python/mqt/qmap/plugins/qiskit.py
denialhaag: The idea was to also have `compile()`, `load_architecture()`, and `load_calibration()` in the `sc` module. I don't think there's a way to have `sc.pyi` in the "package root" and keep the `sc` module. An option would be something like `pysc.pyi` and a `sc` module. 🤔

python/mqt/qmap/sc/compile.py
denialhaag: Ahh yes, I forgot to undo that. I'll do it on Monday!

python/mqt/qmap/plugins/qiskit.py
denialhaag: But maybe it's also a non-issue. I think the duplication is a result of my initial confusion about the `compile()` function being in the package root although it only works for sc circuits, but maybe that's just okay. 🤔

burgholzer: Yeah. there is no way to have the compiled extension be called `sc` and there still be a Python package `sc`.
The `pysc` solution would be fairly similar to the previous `pyqmap` solution.

I think the "proper" solution here would be to also introduce a `plugins` subpackage in QMAP, similar to MQT Core and QCEC that contains all the extra code for Qiskit interoperability. Then we would only have compiled code in the individual submodules.
Eventually, we should look into how we can expose the layout and routing methods of the `mqt.qmap.sc` package as proper transpiler plugins in Qiskit anyway.

Just a couple of thoughts though.

python/mqt/qmap/sc/compile.py
denialhaag: Done. 👍

python/mqt/qmap/plugins/qiskit.py
denialhaag: I have moved the compiled modules back to the package root now. For now, I created a `clifford_synthesis_util.py` and a `sc_util` subpackage. I noticed that there already is a `plugins` subpackage. Is it okay to just move the util functions there? 🤔

python/mqt/qmap/__init__.py
```diff
@@ -45,10 +45,13 @@ def _dll_patch() -> None:
...
+
 from ._version import version as __version__
-from .clifford_synthesis import optimize_clifford, synthesize_clifford
-from .compile import compile  # noqa: A004
-from .subarchitectures import SubarchitectureOrder
+from .clifford_synthesis_util import optimize_clifford, synthesize_clifford
+from .sc_util import (
+    SubarchitectureOrder,
+    compile,  # noqa: A004
+)
```
burgholzer: Let's not reexport these in the top level unit file anymore.
It is fine if people have to include the subpackage in their import.

python/mqt/qmap/clifford_synthesis_util.py
```diff
@@ -15,28 +15,19 @@
...
 
-__all__ = [
-    "optimize_clifford",
-    "synthesize_clifford",
-]
-
-
-def __dir__() -> list[str]:
-    return __all__
-
```
burgholzer: I see no reason for this to be removed.

python/mqt/qmap/clifford_synthesis_util.py
```diff
@@ -6,7 +6,7 @@
 #
 # Licensed under the MIT License
 
-"""Main entry point for the Clifford synthesis module."""
+"""Utilities for Clifford synthesis."""
```
burgholzer: I'd probably move this to "plugins/qiskit/clifford_synthesis".
I am personally not the biggest fan of "utils" files/modules/directories. They tend to become random collections of unrelated things.

Moving these methods into a dedicated place with a dedicated purpose seems like a better idea here (and for the other files).

This here feels particularly wrong because it moves a "main entry point" to a "utils"/"Resterampe"

python/mqt/qmap/plugins/qiskit.py
burgholzer, line 1: It's probably worth to split this up into individual submodules (`plugins/qiskit/sc.py`, `plugins/qiskit/clifford_synthesis.py`, etc.)

python/mqt/qmap/sc.pyi
```diff
@@ -6,7 +6,7 @@
 #
 # Licensed under the MIT License
 
-"""Python bindings module for MQT QMAP."""
+"""Python bindings module for superconducting module."""
```
burgholzer: That sentence doesn't seem to make too much sense 😉

python/mqt/qmap/sc_util/compile.py
```diff
@@ -6,31 +6,16 @@
 #
 # Licensed under the MIT License
 
-"""Main entry point for the compilation module."""
+"""Compile a circuit."""
```
burgholzer: Just a note here. This is ought to be replaced by a proper transpiler plugin eventually.

python/mqt/qmap/sc_util/load_architecture.py
```diff
@@ -6,25 +6,19 @@
 #
 # Licensed under the MIT License
 
-"""Module for loading architectures."""
+"""Load a super-conducting architecture."""
```
burgholzer:
```suggestion
"""Load a superconducting architecture."""
```
Consistency in spelling would be nice. I don't particularly care either way as long as it is consistent.

python/mqt/qmap/sc.pyi
denialhaag: Good point. Fixed.

python/mqt/qmap/plugins/qiskit.py
denialhaag: I moved everything to the `plugins` subpackage. Let me know what you think!

python/mqt/qmap/clifford_synthesis_util.py
denialhaag: Yes, good point! I have moved it to the `plugins` subpackage now.

python/mqt/qmap/clifford_synthesis_util.py
denialhaag: That was a mistake. It's back. 👍

python/mqt/qmap/__init__.py
denialhaag: Good point. Done.

UPGRADING.md
```diff
@@ -4,6 +4,12 @@ This document describes breaking changes and how to upgrade. For a complete list
 
 ## [Unreleased]
 
+With this release, the Python package has been restructured.
+In particular, the `mqt.qmap.pyqmap` module has been discontinued.
+Classes and functions can now be imported from the more descriptive `mqt.qmap.clifford_synthesis`, `mqt.qmap.hybrid_mapper`, `mqt.qmap.na`, and `mqt.qmap.sc` modules.
+The superconduction module's `compile()` function has been moved to `mqt.qmap.plugins.qiskit.sc`.
```
burgholzer:
```suggestion
The superconducting module's `compile()` function has been moved to `mqt.qmap.plugins.qiskit.sc`.
```

burgholzer (approved):
I think this should be good to go for now!
The README snippet still needs an update and there is one small typo.
Feel free to adjust that and then merge this PR.

denialhaag:
> I think this should be good to go for now! The README snippet still needs an update and there is one small typo. Feel free to adjust that and then merge this PR.

Done. Thank you! I set the PR to auto-merge now. 👍

PR #844, opened 2025-11-25
Added checks to support gate types CZ and CY in the Mapper

### LINKED ISSUE #843: 🐛 Not supported control gates in Mapper.cpp
### System and Environment Information

Ubuntu- 22.04 in docker image
C++ version

### Bug Description

We are using a Device supporting CZ as nativ gate while in Mapper.cpp assert(g->getType() == qc::X ) we propse therefore to extend it for cases of CZ and CY else the method crashes and aboards

### Steps to Reproduce

We transpile an arbitrary circuit into an architectures native gate set and try to map it on the couplingmap using Qmap.
Since our gate set contains Cz we need to decompose all Cx operations into HCzH ... . In this case Qmap crashes as it does not handle the case of an operation having controll qubits without being the type of quake.X0P

YouAndre:
## Description
Supplemented the Assert  X operation with the checks also for Cz and Cy operations to support circuits using these operations instead of Cx.

Fixes #843  

## Checklist:

- [ ] I have added appropriate tests that cover the new/changed functionality.
- [ ] I have updated the documentation to reflect these changes.
- [ ] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.
- [ ] I have added migration instructions to the upgrade guide (if needed).
- [ ] The changes are fully tested and pass the CI checks.

### DIFF
--- a/src/sc/Mapper.cpp
+++ b/src/sc/Mapper.cpp
@@ -557,7 +557,8 @@ void Mapper::countGates(decltype(qcMapped.cbegin()) it,
           info.totalLogFidelity += architecture->getSingleQubitFidelityCost(q1);
         }
       } else {
-        assert(g->getType() == qc::X);
+        assert(g->getType() == qc::X || g->getType() == qc::Y ||
+               g->getType() == qc::Z);
         ++info.cnots;
         ++info.gates;
         auto q1 =


ystade:
@YouAndre just because you asked me to review your PR here and for cross-referencing, you have seen @burgholzer's comment in #843, right?

PR #874, opened 2025-12-09
👷 Update C++ test matrices

### LINKED ISSUE #703: 🐛 Python tests fail on macOS 15
### System Information

- Operating system: macOS 15
- MQT QMAP version: 3.2.0
- Python versions: 3.9, 3.10, 3.11, 3.12, 3.13 

### Bug Description

The Python tests with minimum versions currently fail on macOS 15, see, for example, [this workflow run](https://github.com/munich-quantum-toolkit/qmap/actions/runs/16353780901/job/46208198937).

### Steps to Reproduce

The error occurs in the CI.

### See Also

- https://github.com/munich-quantum-toolkit/ddsim/issues/597
- https://github.com/munich-quantum-toolkit/qcec/issues/653

denialhaag:
## Description

This PR updates the C++ test matrices:

- We stop testing on the `ubuntu-22.04` and `ubuntu-22.04-arm` runners
- We stop testing `clang-19` builds and instead start testing `clang-21` builds

Furthermore, this PR removes Python 3.13t builds from the CD matrix due to their experimental nature.

Fixes #703

## Checklist:

- [x] ~I have added appropriate tests that cover the new/changed functionality.~
- [x] ~I have updated the documentation to reflect these changes.~
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,18 @@ _If you are upgrading: please see [`UPGRADING.md`](UPGRADING.md#unreleased)._
 
 ### Changed
 
+- 👷 Stop testing on `ubuntu-22.04` and `ubuntu-22.04-arm` runners ([#874]) ([**@denialhaag**])
+- 👷 Stop testing with `clang-19` and start testing with `clang-21` ([#874]) ([**@denialhaag**])
+- 👷 Fix macOS tests with Homebrew Clang via new `munich-quantum-toolkit/workflows` version ([#874]) ([**@denialhaag**])
+- 👷 Re-enable macOS tests with GCC by disabling module scanning ([#874]) ([**@denialhaag**])
 - ✨ Enable code generation for relaxed routing constraints ([#848]) ([**@ystade**])
 - ✨ Add `max_filling_factor` to scheduler in Zoned Neutral Atom Compiler ([#847]) ([**@ystade**])
 - ✨ Added extension to the hybrid routing mapper to also support Bridge gates, Passby moves and Flying ancillas ([#832]) ([**@lsschmid**])
 - ✨ Added hybrid synthesis routing for iterative circuit constructions ([#832]) ([**@lsschmid**])
+
+### Removed
+
+- 🔥 Remove wheel builds for Python 3.13t ([#874]) ([**@denialhaag**])
 
 ## [3.4.0] - 2025-10-15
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -109,3 +109,15 @@ else()
...
+# https://www.reddit.com/r/cpp_questions/comments/1kwlkom/comment/ni5angh/
+if(CMAKE_SYSTEM_NAME STREQUAL "Darwin" AND CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.28")
+    set(CMAKE_CXX_SCAN_FOR_MODULES OFF)
+  else()
+    message(WARNING "CMake 3.28+ is required to disable C++ module scanning on macOS with GCC. "
+                    "Current version: ${CMAKE_VERSION}. "
+                    "Consider upgrading CMake to avoid potential build issues.")
+  endif()
+endif()
```
denialhaag: @burgholzer, do you know why this doesn't fix the `macOS` x `gcc` tests in this case? Is this not supposed to work in the root `CMakeLists.txt`? I have only put it there for testing purposes now, as QMAP should inherit the setting from Core, if I'm not mistaken. 🤔

denialhaag: We have the exact same issue in QCEC.

burgholzer: Could you try with a recent MQT Core rev in `cmake/ExternalDependencies` that includes the fix in MQT Core?

Another potential solution: You may need to set this _before_ including the project source code as well as the external dependency. Try moving it to the top here.

denialhaag: Thanks! I will try later tonight so as not to clog up the CI now. 🤠

denialhaag: It looks like the latter has done the trick!

Do we want to leave this in for now and then remove it once we update to the new version of MQT Core? 🤔

burgholzer: Seems reasonable to me. Then we can get this in now.
Just has the slight danger of us forgetting about it.

denialhaag: Perfect, I'll update the changelog and upgrade guide tomorrow!

I guess another option would be removing it for now and creating an issue that macOS x GCC builds fail. Seems a bit forced, though. 🤔

burgholzer: Nah. Let's merge this tomorrow. It does actually fix build errors in the library. So it's a valid fix.

PR #1059, opened 2026-07-03
✨Add Photonic Subcircuit Compiler

tobi-forster:
🤖 *AI text below* 🤖

Follow-up at `580ee7f9`: Codex assisted with the routing and propagation fixes, regression tests, and review replies. The `photonics` extra now requires only `torch>=2.9.0`; NumPy remains in the separate evaluation dependency group.

## Description

This PR introduces `mqt.qmap.ph`, a new Python subpackage implementing a photonic subcircuit compiler for MZI-mesh chips. The compiler routes photons to the computation zone by finding the lowest-loss path through the chip's MZI mesh, then fits the phase-shifter parameters to a target unitary via PyTorch gradient descent. It returns the phases to program and the input/output ports to use — no hardware simulation is needed to compile.

A separate, non-installed evaluation harness under `eval/ph` reproduces the paper's benchmark: it simulates a compiled circuit with Perceval, reporting coincidence rate and TVD against the ideal output distribution, and compares it to a fixed-placement baseline to verify routing improves performance under transmission loss.

Dependencies:
- **Compiler** (`mqt.qmap[photonics]`): numpy ≥ 1.26.0, torch ≥ 2.9.0
- **Evaluation only** (`eval/ph`, `evaluation` dependency group, not shipped with the package): numpy, pandas ≥ 2.3.3, perceval-quandela ≥ 1.1.0, torch

## Checklist

- [x] I have added appropriate tests that cover the new/changed functionality.
- [x] I have updated the documentation to reflect these changes.
- [x] I have added entries to the changelog for any noteworthy additions, changes, fixes, or removals.

**If PR contains AI-assisted content:**

- [x] I have disclosed the use of AI tools in the PR description as per our [AI Usage Guidelines](https://github.com/munich-quantum-toolkit/qmap/blob/main/docs/ai_usage.md).
- [x] AI-assisted commits include an `Assisted-by: [Model Name] via [Tool Name]` footer.
- [x] I confirm that I have personally reviewed and understood all AI-generated content, and accept full responsibility for it.


tobi-forster:
@coderabbitai full review

tobi-forster:
@coderabbitai full review

tobi-forster:
@coderabbitai full review

python/mqt/qmap/ph/graph.py
```diff
@@ -0,0 +1,558 @@
...
+    """
+    cache = np.sqrt((1 - r[0]) * r[1]) + np.sqrt(r[0] * (1 - r[1]))
+    return cache**2
+
+
+def generate_beam_splitter_matrix(
+    chip_size: int,
+    ideal_bs: bool = False,
+    rng: np.random.Generator | None = None,
+) -> np.ndarray:
```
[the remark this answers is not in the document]
tobi-forster: @coderabbitai FBT001 and FBT002 is not activated for this repository, is it?

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,535 @@
...
+        rng=noise_rng,
+    )
+    baseline_virtual_chip = create_mzi_chip(
+        beam_splitter_reflectivities,
+        baseline_phase_shifter_params_2d,
+        phase_error=phase_error,
+        chip_size=chip_dim,
+        exclude_edge_phase_shifters=config.exclude_edge_phase_shifters,
+        rng=noise_rng,
+    )
```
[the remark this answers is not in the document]
tobi-forster: @coderabbitai Since the phase noise is not a constant offset but a random factor influenced by environmental conditions, this is not sensible.

tobi-forster:
@coderabbitai full review

python/mqt/qmap/ph/graph.py
```diff
@@ -0,0 +1,563 @@
...
+
+    # Final exact mean/std normalisation.
+    current_mean = np.mean(bs_values)
+    current_std = np.std(bs_values)
+    if current_std > 0:
+        bs_values = (bs_values - current_mean) * (target_std / current_std) + target_mean
+    else:
+        bs_values[:] = target_mean
+
+    return bs_values
```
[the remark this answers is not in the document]
tobi-forster: @coderabbitai Are PLR9012 and PLR9015 actually activated in this repository?

eval/ph/perceval_simulation.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def create_mzi_chip(
+    bs_list: np.ndarray,
+    ps_matrix: torch.Tensor | np.ndarray,
+    phase_error: float | None,
+    chip_size: int,
+    exclude_edge_phase_shifters: bool = False,
+    rng: np.random.Generator | int | None = None,
+) -> pcvl.Circuit:
```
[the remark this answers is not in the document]
tobi-forster: @coderabbitai FBT is not enabled, right?

ystade:
@coderabbitai There are some unresolved comments. Please, check whether they were already tackled and then resolve them, otherwise comment, why they are not resolved yet.

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ _If you are upgrading: please see [`UPGRADING.md`](UPGRADING.md#380)._
 
 ### Changed
 
+- 🚀 Add photonic subcircuit compiler that optimizes coincidence rates ([#1059])
```
ystade: Nitpick: I do not know the meaning of the rocket here. We take the symbols with their semantic meaning from gitmoji.dev. Hence, :tada:, is most fitting here imho.

ystade: You listed that change under the already released version. Instead, add the subsection `## Changed` for the `# Unreleased` version and move this item there.

noxfile.py
```diff
@@ -87,13 +88,26 @@ def _run_tests(
...
         env=env,
     )
+    # === TEMPORARY (macOS-Intel torch wheel gap) — remove once resolved ===
+    # torch >= 2.3 ships no macOS x86_64 wheels, so the `photonics` extra (which
+    # pulls in torch) cannot be installed on Intel macOS runners. Skip just that
+    # extra there; the photonics tests self-skip via `pytest.importorskip`, and
+    # the torch-free tests (e.g. test_graph.py) still run. Remove this block and
+    # restore the unconditional `"--extra", "photonics"` args below once torch
+    # publishes macOS x86_64 wheels again or Intel macOS runners leave the matrix.
+    photonics_extra: tuple[str, ...] = ("--extra", "photonics")
```
ystade: I am wondering whether `photonics_extra` there is a more meaningful name for this variable. (1) one could imagine to list other extensions than photonics in the future; (2) that is an entirely "me"-problem: I did not know that these extras officially called extras since I haven't worked with Python so much recently. However, sometimes, they are also called `optional_dependencies`, which in my opinion is a more expressive variable name and better suited here.

noxfile.py
```diff
@@ -87,13 +88,26 @@ def _run_tests(
         *install_args,
         env=env,
     )
+    # === TEMPORARY (macOS-Intel torch wheel gap) — remove once resolved ===
+    # torch >= 2.3 ships no macOS x86_64 wheels, so the `photonics` extra (which
+    # pulls in torch) cannot be installed on Intel macOS runners. Skip just that
+    # extra there; the photonics tests self-skip via `pytest.importorskip`, and
+    # the torch-free tests (e.g. test_graph.py) still run. Remove this block and
+    # restore the unconditional `"--extra", "photonics"` args below once torch
+    # publishes macOS x86_64 wheels again or Intel macOS runners leave the matrix.
```
ystade: @burgholzer For me this solution is fine, just tagging you here for awareness when you review it.

pyproject.toml
```diff
@@ -405,6 +417,9 @@ docs = [
   "sphinxext-opengraph>=0.13.0",
   "walkerlayout>=1.0.2",
   "qiskit[qasm3-import,visualization]>=1.0.0",
+  # The photonic subcircuit compiler doc executes compile_subcircuit, which needs
+  # torch (but not the perceval/pandas simulation stack — those cells are not executed).
```
ystade: Having docs here is nice but since no other packages have a docstring, I would leave these two lines away also here.

pyproject.toml
```diff
@@ -379,6 +385,12 @@ exclude = [
     "eval/**",
 ]
 
+[[tool.ty.overrides]]
```
ystade, line 381: I am wondering whether this is really needed. Yes, I agree that it is nice to lint all the files without having to install `torch`. However, this must be installed anyways for the tests. So, if it happens earlier or later does not matter, does it?

docs/ph_subcircuit_compiler.md
ystade, line 1: This file is not added to the menu tree of the main documentation page but should. Probably, one also wants to add a bullet in the main text announcing the PH compiler.
And, additionally, not really related to this file, one wants to add the corresponding publication to the list of publications. This has the advantage that it can directly be referenced here.

docs/ph_subcircuit_compiler.md
```diff
@@ -0,0 +1,235 @@
...
+pip install "mqt.qmap[photonics]"
+```
+
+:::
+
+## Hardware Model
+
+The chip is a staggered MZI mesh with `chip_dim` spatial modes and `chip_dim`
+MZI layers. Layers alternate between _full_ layers (MZIs coupling modes 0–1,
+2–3, …) and _half_ layers (MZIs coupling modes 1–2, 3–4, …). Each MZI consists
```
ystade: It is not really _half_, is it? It's just one coupling less, or?

eval/ph/results/pipeline_results.csv
ystade, line 1: I do not have a big objection having results here, but at the same time also no need to have the results here. The sole purpose of the script is that everyone can reproduce the results (at the same time, I also got e-mails, asking for the original results, so perhaps it is good to have them here as a reference). Is there anything bad about having them here?

docs/ph_subcircuit_compiler.md
```diff
@@ -0,0 +1,235 @@
...
+)
+```
+
+### Inspect the compiled result
+
+The result carries everything needed to drive the chip:
+
+```{code-cell} ipython3
+print("Phase-shifter settings (shape):", tuple(result.phases.shape))  # (chip_dim, chip_dim)
+print("Inject photons in input state :", result.input_ports)          # length chip_dim (dual-rail)
```
ystade: I find that a little irritating, that the input uses a kind of one-hot encoding and the outputs list the indices. Can this be streamlined?

eval/ph/data_collection.py
```diff
@@ -0,0 +1,452 @@
...
+# SPDX-License-Identifier: MIT
+#
+# Licensed under the MIT License
+
+"""Batch data collection and aggregation for the photonic compiler pipeline."""
+
+# ------------------------------------------------------------------------------
+# Setup used for QCE26 paper submission.
+# Future structure of the code will focus on the compilation,
+# not the simulation and data collection.
```
ystade: I do not understand this comment. Would it be enough to say that the script is for reproducing the results of the paper?

eval/ph/data_collection.py
```diff
@@ -0,0 +1,452 @@
...
+        target_dims_list: Candidate target unitary dimensions.
+
+    Returns:
+        List of valid :class:`Setup` instances.
+    """
+    setups = []
+    for num_modes, target_dim in product(num_modes_list, target_dims_list):
+        if target_dim > num_modes:
+            continue
+        if num_modes % 2 != 0 or target_dim % 2 != 0:
```
ystade: Not super important, that bothers my CS heart: Why are these lists not filtered in the first place, i.e., only contain even numbers? Without the docstring, I would not expect this kind of filtering from the function (besides that it is inefficient).

eval/ph/data_collection.py
```diff
@@ -0,0 +1,452 @@
...
+    num_modes: int,
+    kind: str,
+    hardware_data_dir: pathlib.Path,
+    rng: np.random.Generator,
+) -> np.ndarray:
+    """Return per-mode transmission coefficients, preferring the paper's data.
+
+    If ``hardware_data_dir`` contains ``{num_modes}_mode_{kind}_transmissions.txt``
+    the exact paper values are loaded (used as-is; the files are already
+    normalised so the maximum is 1.0).  If no such file exists, per-mode
```
ystade: This is counter-intuitive to me. I would expect the following: the `hardware_data_dir` can also be `None` in which case random data is used, however, if it is set to some value and the file does not exist, the script terminates with an error.

eval/ph/data_collection.py
```diff
@@ -0,0 +1,452 @@
...
+    custom_bs_data: dict[int, np.ndarray] | None,
+    hardware_data_dir: pathlib.Path = _HARDWARE_DATA_DIR,
+) -> dict[int, tuple[np.ndarray, np.ndarray, np.ndarray]]:
+    """Build per-num_modes hardware parameter cache.
+
+    Args:
+        setups: All setups whose unique ``num_modes`` values are cached.
+        base_seed: Base RNG seed; the seed for ``num_modes`` is
+            ``base_seed + 10 * num_modes``.
+        input_losses: When ``True``, load the paper's exact input transmissions
```
ystade: Why not a file path as above?

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,537 @@
...
+@dataclass
+class OptimizationConfig:
+    """Hyperparameters for the phase-shifter optimisation.
+
+    Attributes:
+        lr: Initial Adam learning rate.
+        threshold: Fidelity-loss value below which optimisation terminates
+            early.
+        max_iterations: Maximum gradient steps.
+        exclude_edge_phase_shifters: If ``True``, the two corner phase
```
ystade: edge or corner? ;)

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,537 @@
...
+    exclude_edge_phase_shifters: bool = False
+    optimize_routing_parameters: bool = True
+
+
+@dataclass
+class CompilationResult:
+    """Output of a single :func:`compile_subcircuit` call.
+
+    This is the end-user result of compiling a target unitary onto a physical
+    chip.  It carries everything needed to drive the hardware: the phase-shifter
```
ystade: One space too much.

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,537 @@
...
+        input_ports: Chip-wide binary input-state vector of length ``chip_dim``
+            marking the modes into which photons are injected (dual-rail).
+        output_ports: Physical mode indices of the computation zone where the
+            output photons are measured.
+        loss: Final fidelity loss of the proposed compiler optimisation.
+        compute_time: Wall-clock seconds for the proposed compiler (routing
+            + optimisation).
+    """
+
+    phases: torch.Tensor
```
ystade: Is there a good reason why the tensor type and in the following quite basic types are used? Wouldn't it be more consistent to use basic types throughout?

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,537 @@
...
+        baseline_performance: Same metrics for the baseline strategy.
+        loss: Final fidelity loss of the proposed compiler optimisation.
+        baseline_loss: Final fidelity loss of the baseline optimisation.
+        compute_time: Wall-clock seconds for the proposed compiler (routing
+            + optimisation).
+        baseline_compute_time: Wall-clock seconds for the baseline
+            optimisation.
+    """
+
+    performance: dict[str, Any]
```
ystade: Just an idea, would it make the structure more concise, if one has one class with performance, loss, and compute time, and then another class which would have the former as two attributes for the baseline and for the real run?

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,537 @@
...
+    target_unitary_opt: torch.Tensor,
+    beam_splitter_reflectivities: np.ndarray,
+    movement_mask: torch.Tensor,
+    config: OptimizationConfig,
+    chip_dim: int,
+    input_ports: list[int],
+    active_cols_computation_zone: list[int],
+    output_ports: list[int],
+) -> tuple[float, torch.Tensor]:
+    """Optimize phase-shifter parameters for the proposed routing path.
```
ystade, line 250: You used BrE throughout. I did not comment on it, but will do now, as here you are using AmE. Usually, our docstrings use AmE. I am a big fan of consistency, so I would prefer one consistent choice for the PH project (and preferably across QMAP).

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,537 @@
...
+        input_ports=converted_input_ports,
+        output_ports=output_ports,
+        loss=losses,
+        compute_time=proposed_compute_time,
+    )
+
+
+def evaluate_subcircuit(
+    compilation: CompilationResult,
+    beam_splitter_reflectivities: np.ndarray,
```
ystade: I am just wondering here, whether there is a good reason that we use `torch.Tensor`, `np.ndarray`, and basic types. Is there, or could one get rid of one type?

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,537 @@
...
+    """Evaluate a compiled subcircuit against the baseline via Perceval simulation.
+
+    Takes the output of :func:`compile_subcircuit` and simulates the resulting
+    chip, comparing it to a fixed dual-rail baseline placement (first
+    ``target_dim`` modes, no routing).  Both are simulated on the same hardware
+    model (beam-splitter imperfections, phase noise, input/output transmission
+    losses) and their output distributions are compared to an ideal Perceval
+    simulation.
+
+    This step requires a classical simulation of the chip and is intended for
```
ystade: Very nitpicky, I think there is a label to display something as a `Note:`. This might fit nicely here for this paragraph.

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,537 @@
...
+
+    # When a seed is supplied, share one generator across both create_mzi_chip
+    # calls so the evaluation is reproducible; ``None`` preserves the original
+    # behaviour of drawing fresh entropy per chip.
+    noise_rng = None if phase_noise_seed is None else np.random.default_rng(phase_noise_seed)
+
+    baseline_active_cols = get_baseline_active_cols(target_dim)
+    baseline_input_ports = get_baseline_input_ports(baseline_active_cols, chip_dim)
+
+    beam_splitter_reflectivities_t = torch.as_tensor(beam_splitter_reflectivities, dtype=torch.float64)
```
ystade: You might already realized that I am a big fan of consistency... I wonder whether the `_t` suffix is according to the naming convention used throughout this project. I do not think it is particularily bad, I just want to ensure that it is consistently used. Less fitting but might also apply to the `_2d` three lines below.

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,537 @@
...
+        beam_splitter_reflectivities_t,
+        config,
+        chip_dim,
+        baseline_active_cols,
+    )
+    baseline_compute_time = time.time() - baseline_start
+
+    ideal_prob_dist, baseline_ideal_prob_dist = _compute_ideal_distributions(target_unitary, target_dim)
+
+    virtual_chip = create_mzi_chip(
```
ystade: The following is very redundant for the baseline and the proposed method. You suggest having one evaluation function that is once called for the baseline and once for the proposed method. Does that make sense?

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,537 @@
...
+) -> CompilationResult:
+    """Compile a target unitary onto the chip and return the phases to program.
+
+    ``chip_dim`` is derived from ``len(input_transmissions)`` and
+    ``target_dim`` from ``target_unitary.shape[0]``.
+
+    The compiler searches for the optimal photon routing through the chip
+    (the path minimising overall photon loss, independent of the target
+    operation) and then optimises the phase-shifter parameters for that
+    placement.  No hardware simulation is performed, so this step is suitable
```
ystade: Double space.

ystade: The last sentence could also be rendered as a note if you like.

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,537 @@
...
+        beam_splitter_reflectivities: 1D array of chip beam-splitter
+            reflectivities as produced by
+            :func:`graph.generate_beam_splitter_matrix`.
+        input_transmissions: Per-mode input transmission coefficients, shape
+            ``(chip_dim,)``.  Its length determines ``chip_dim``.
+        output_transmissions: Per-mode output transmission coefficients, shape
+            ``(chip_dim,)``.
+        target_unitary: Target unitary tensor of shape ``(target_dim, target_dim)``.
+            Its first dimension determines ``target_dim``.
+        config: Optimisation hyperparameters.  Defaults to
```
ystade: Double space. I'd recommend search&replace.

python/mqt/qmap/ph/subcircuit_compilation.py
ystade: Also the code uses AmE.

python/mqt/qmap/ph/baseline.py
```diff
@@ -0,0 +1,68 @@
...
+    Args:
+        target_dim: Dimension of the target unitary.
+
+    Returns:
+        List of even indices ``[0, 2, 4, …, target_dim - 2]``.
+    """
+    return [i for i in range(target_dim) if i % 2 == 0]
+
+
+def get_baseline_input_ports(active_cols: list[int], chip_dim: int) -> list[int]:
```
ystade: I commented already elsewhere. Would it make sense to also use a list of indices here instead of kinda one-hot enconding?

python/mqt/qmap/ph/graph.py
```diff
@@ -0,0 +1,566 @@
...
+
+
+def _edge_cost_from_mzi_block(
+    fidelity_list: list[float],
+    fidelity_offset: int,
+    source_node_idx: int,
+    mzi_count_in_layer: int,
+    num_parallel_photons: int,
+) -> float:
+    """Return ``-log(product of fidelities)`` over the MZI block for one edge.
```
ystade: It feels like this function is doing two things at a time:
1. Extracting the fidelity at a certain location
2. Converting it with the formula to the edge cost.
I'd suggest splitting this into two functions according to the purposes above.

python/mqt/qmap/ph/graph.py
```diff
@@ -0,0 +1,566 @@
...
+        target_dim: Dimension of the target unitary.
+        input_transmission: Per-mode input transmission coefficients, shape
+            ``(chip_dim,)``.
+        output_transmission: Per-mode output transmission coefficients, shape
+            ``(chip_dim,)``.
+        beam_splitter_reflectivities: 1D array of beam splitter reflectivities
+            as produced by :func:`generate_beam_splitter_matrix`.
+
+    Returns:
+        A tuple ``(graph, pos, layers)`` where *graph* is the weighted directed
```
ystade: I find this type a little suboptimal. Imho, the cleanest solution would be to wrap this data into its own type and then have corresponding accessors for the x,y coordinates and the layers. Depending on your motivation, feel free to ignore this and just resolve it.

python/mqt/qmap/ph/perceval_simulation.py
```diff
@@ -0,0 +1,240 @@
...
+            When provided, a loss channel ``LC(1 - t)`` is appended to each
+            mode.
+
+    Returns:
+        A tuple ``(processor, probability_distribution)`` where *processor*
+        is the configured :class:`perceval.Processor` and
+        *probability_distribution* is the raw BSDistribution mapping output
+        states to probabilities.
+    """
+    processor = pcvl.Processor("SLOS", chip_dim)
```
ystade: Probably, it's all right. I just saw "SLOS" already in another file and was wondering, whether this option is correctly pushed forward to its destination. Please, check but feel free to resolve, if this is all intended.

python/mqt/qmap/ph/routing.py
```diff
@@ -0,0 +1,348 @@
...
+
+    path.reverse()
+
+    if len(path) == 1 and predecessors[target_node] is None:
+        return []
+
+    return path
+
+
+def get_best_route(
```
ystade, line 34: I was wondering whether we actually need the dependy `rustworkx` if we have our won shortest path algorithm anyways. That would be one depency less than. Is this possible?

python/mqt/qmap/ph/routing_to_phases.py
```diff
@@ -0,0 +1,171 @@
...
+
+    return effective_params, grad_mask, refined_mask
+
+
+def reshape_flattened_params_to_grid(
+    params_1d: torch.Tensor,
+    num_modes: int,
+    exclude_edge_phase_shifters: bool = False,
+) -> torch.Tensor:
+    """Inflate a 1D parameter vector into a 2D phase-shifter grid.
```
ystade, line 231: Do I understand correctly that these conversions are necessary, because the interface of the optimizer we are using accepts lists and at the end we need a matrix? There is not way to avoid these transformations, right?

python/mqt/qmap/ph/unitary_to_phase_compilation.py
```diff
@@ -0,0 +1,625 @@
...
+def fidelity_loss(
+    effective_unitary: torch.Tensor,
+    target_unitary: torch.Tensor,
+    active_cols: list[int] | None = None,
+    active_cols_target: list[int] | None = None,
+    baseline_outputs: list[int] | None = None,
+) -> torch.Tensor:
+    """Compute the normalised fidelity loss between two unitaries.
+
+    Loss is defined as ``1 - |Tr(U_tgt† @ U_eff)|² / N²``, where *N* is
```
ystade: This probably works in most cases but nevertheless, I would like to avoid having unicode characters in source files. Can you check, whether there is a way to embedd latex code into docstrings?

python/mqt/qmap/ph/unitary_to_phase_compilation.py
```diff
@@ -0,0 +1,625 @@
...
+            shifters are excluded from the parameter set.
+        optimize_routing_parameters: If ``True``, routing cells contribute a
+            single trainable degree of freedom.
+        early_stop_patience: Number of consecutive steps without improvement
+            before optimisation is terminated early.
+        min_improvement: Minimum absolute loss decrease required to reset the
+            patience counter.
+
+    Returns:
+        A dictionary with the following keys:
```
ystade: Perhaps, this is the pythonic way to do things, if so, please ignore the following. Me, coming from C++, I would rather define a plain data class that has the keys as attributes and would return an object of that class.

test/python/ph/conftest.py
```diff
@@ -0,0 +1,71 @@
...
+    ("mqt", _PYTHON_SRC / "mqt"),
+    ("mqt.qmap", _PYTHON_SRC / "mqt" / "qmap"),
+]:
+    if _pkg not in sys.modules:
+        _mod = ModuleType(_pkg)
+        _mod.__path__ = [str(_path)]
+        _mod.__package__ = _pkg
+        sys.modules[_pkg] = _mod
+
+sys.path.insert(0, str(_PYTHON_SRC))
```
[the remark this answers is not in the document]
ystade: @coderabbitai @tobi-forster Is it actually really necessary to stub mqt and qmt.qmap. Yes, indeed, the photonics compiler does not require a C++ build but we are building the C++ code anyways in the CI so I di not understand what one gains through stubbing it.

eval/ph/tests/test_baseline.py
```diff
@@ -0,0 +1,118 @@
...
+        expected = torch.eye(4, dtype=torch.complex128)
+        assert torch.allclose(result, expected)
+
+    @staticmethod
+    def test_target_block_is_correctly_placed() -> None:
+        """Test that the target unitary values appear in the top-left block."""
+        u = np.array([[1 + 2j, 3 + 4j], [5 + 6j, 7 + 8j]])
+        result = embed_target_unitary_into_chip(u, chip_dim=4, target_dim=2)
+
+        assert result[0, 0] == pytest.approx(1 + 2j)
```
ystade, line 68: Why only approximate? Because complex numbers are stored as floats internally?

test/python/ph/test_data_collection.py
```diff
@@ -0,0 +1,141 @@
...
+from mqt.qmap.ph.subcircuit_compilation import OptimizationConfig
+
+
+class TestBuildSetupGrid:
+    """Tests for build_setup_grid."""
+
+    @staticmethod
+    def test_filters_target_larger_than_chip() -> None:
+        """Test that setups where target_dim > num_modes are excluded."""
+        setups = build_setup_grid([4], [6])
```
ystade: Shouldn't this function rather raise an exception on illegal arguments?

test/python/ph/test_routing.py
```diff
@@ -0,0 +1,265 @@
...
+        later layer) must span exactly four modes, not two.
+        """
+        mask = route_to_movement_mask([0, 0, 0, 0, 1, 1, 0], chip_dim=8, target_dim=4)
+        # Cross starts at mode 0 and is target_dim (=4) rows wide on chip layer 2.
+        assert torch.all(mask[0:4, 2] == MaskState.CROSS)
+        # Modes outside that target_dim-wide window are untouched on that layer.
+        assert torch.all(mask[4:8, 2] == MaskState.BAR)
+
+
+class TestGetBestRoute:
```
ystade, line 149: I do not know where to comment that, I'll just do it here. All the tests form a very sophisticated test suite. Most of them test small details of the implementation (which is fine and also necessary). However, I am missing (or I have overlooked it) an end-to-end test that feeds in a certain phase matrix (or multiple to test multiple scenarios) and tests where the outcome is as expected. To this end, the input phase matrix should deterministically lead to one certain output.

This is more of a fundamental comment: In C++ one has a better separation of the public interface and functions that are only called internally. While it is quite useful to also test the internal functions from time to time, to verify that the internals work as expected, most of the test should through the user facing interface, i.e., calling the functions a user would call directly with a certain input and then verifying that the output is as expected. This allows in particular, to change the internals completely as long as the public facing interface remains the same.

ystade (dismissed):
@tobi-forster I am very happy to see this kick-start of a compiler for photonics. Really, great work that resulted in quite some lines. Hence, I do not recall in detail anymore what all my comments were about but I do recall that there was nothing critical and mostly the comments are more of a remark. Feel free to ask any questions if anything is unclear and (opposed to the coding guidelines) please resolve every comment that is fixed. Otherwise I loose the overview.

Regarding the tests that are skipped: These are skipped because this PR does not touch the C++ code base, so this is all fine.

eval/ph/tests/test_baseline.py
tobi-forster: Yes. For this specific test, one could also switch to an exact check, since small integers can be represented exactly. However, I would leave it like this for generality.

eval/ph/results/pipeline_results.csv
tobi-forster: I think it makes sense to remove them and let everyone interested reproduce the results.

pyproject.toml
tobi-forster: @coderabbitai What is your opinion on the above?

eval/ph/data_collection.py
```diff
@@ -0,0 +1,478 @@
+# Copyright (c) 2023 - 2026 Chair for Design Automation, TUM
```
burgholzer, line 1: 🤖 *AI text below* 🤖

This directory is the appropriate home for the paper evaluation. Please gather all evaluation-only functionality here, including the baseline, Perceval simulation, synthetic hardware model, and evaluation result types. A short README or equivalent documentation here should explain the setup, dependencies, and reproduction command. The evaluation code itself can remain pragmatic; the important point is that it is clearly separated from the installed compiler package.

pyproject.toml
```diff
@@ -67,6 +67,12 @@ visualization = [
     "walkerlayout>=1.0.2",
     "ipywidgets>=8.1.7",
 ]
```
burgholzer, line 70: 🤖 *AI text below* 🤖

Please keep this extra limited to dependencies required by the installed compiler. Pandas is used by data collection, and Perceval is used by simulation and evaluation, so both should move to the evaluation setup rather than being installed with `mqt.qmap.ph`. PyTorch can remain here if the compiler itself continues to require it. This split should also remove some of the lint and CI special cases introduced by the combined dependency set.

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,573 @@
...
+    """
+
+    phases: list[float]
+    input_ports: list[int]
+    output_ports: list[int]
+    loss: float
+    compute_time: float
+
+
+@dataclass
```
burgholzer: 🤖 *AI text below* 🤖

`RunMetrics` and `RunResult` describe the comparison between the proposed method and the experimental baseline; they are not compiler results. Please move them to the evaluation code. The installed API should expose the result needed to program the hardware, without also exposing the structure of the paper evaluation.

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,573 @@
...
+        # Flatten column by column (layer by layer): phase at mode r, layer c -> index c * chip_dim + r.
+        phases=phase_shifter_params_including_routing.t().flatten().tolist(),
+        input_ports=input_ports,
+        output_ports=output_ports,
+        loss=losses,
+        compute_time=proposed_compute_time,
+    )
+
+
+def evaluate_subcircuit(
```
burgholzer: 🤖 *AI text below* 🤖

Please move this evaluation path to `eval/ph`. It runs a second baseline optimization, constructs simulated chips, invokes Perceval, and computes comparison metrics. None of that is required to compile a subcircuit or drive the hardware. Keeping it in the package makes the compiler API and dependency boundary much larger than necessary.

python/mqt/qmap/ph/graph.py
```diff
@@ -0,0 +1,601 @@
...
+
+    Returns:
+        Fidelity value in ``[0, 1]`` indicating how well the MZI swaps
+        the two input modes.
+    """
+    cache = np.sqrt((1 - r[0]) * r[1]) + np.sqrt(r[0] * (1 - r[1]))
+    return cache**2
+
+
+def generate_beam_splitter_matrix(
```
burgholzer: 🤖 *AI text below* 🤖

This is a synthetic hardware model with hard-coded statistics for the evaluation. The compiler already accepts measured beam-splitter reflectivities, so it does not need to generate them. Please move this function to `eval/ph` and document the origin and purpose of these constants there.

python/mqt/qmap/ph/routing.py
```diff
@@ -0,0 +1,337 @@
...
+    """
+
+    MZI = 0  # Compute: both phases are learnable
+    BAR = 1  # Routing: passes light straight through (0, pi)
+    CROSS = 2  # Routing: swaps ports (0, 0)
+    TOP_ONLY = 3  # Virtual PS: top is param, bottom is top + pi
+    BOT_ONLY = 4  # Virtual PS: bottom is param, top is bottom + pi
+
+
+def find_optimal_routing_dag(
```
burgholzer: 🤖 *AI text below* 🤖

This reimplements shortest-path search and path reconstruction even though the graph is already a rustworkx `PyDiGraph`, and rustworkx is an existing QMAP dependency. Please replace this and `reconstruct_path` with `rustworkx.digraph_dijkstra_shortest_paths`. The library implementation should contain only the photonics-specific routing model, not another general graph algorithm.

python/mqt/qmap/ph/graph.py
```diff
@@ -0,0 +1,601 @@
...
+    Args:
+        fidelity: Fidelity value in ``[0, 1]``.
+
+    Returns:
+        Non-negative edge cost ``-log(fidelity)``.
+    """
+    return -np.log(fidelity)
+
+
+def get_edge_fidelity_even_graph_layer(
```
burgholzer: 🤖 *AI text below* 🤖

The even- and odd-layer versions duplicate almost all validation, offset calculation, fidelity selection, and cost conversion. The material difference is the number of MZIs in the corresponding chip layer. Please replace both functions with one edge-cost function that derives this value from the chip-layer parity.

python/mqt/qmap/ph/graph.py
```diff
@@ -0,0 +1,601 @@
...
+        graph: Weighted directed acyclic graph encoding the routing decisions
+            as a shortest-path problem.
+        positions: Maps each node index to an ``(x, y)`` visualization
+            coordinate.  Not used by the routing itself -- retained for
+            plotting and debugging the graph.
+        layers: Per-layer node index arrays as returned by rustworkx.
+    """
+
+    graph: rx.PyDiGraph
+    positions: dict[int, tuple[float, float]]
```
burgholzer: 🤖 *AI text below* 🤖

No caller in this PR uses these plotting coordinates, yet every graph constructs and returns them. Please remove `positions` and the coordinate-generation block. If a graph visualization is added later, it can own the layout it needs.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
```diff
@@ -0,0 +1,643 @@
...
+    mat = torch.eye(num_modes, dtype=torch.complex128)
+    r = torch.as_tensor(reflectivity, dtype=torch.float64, device=mat.device)
+    mat[mode, mode] = torch.sqrt(r)
+    mat[mode, mode + 1] = 1j * torch.sqrt(1 - r)
+    mat[mode + 1, mode] = 1j * torch.sqrt(1 - r)
+    mat[mode + 1, mode + 1] = torch.sqrt(r)
+    return mat
+
+
+def build_unitary_from_components(
```
burgholzer: 🤖 *AI text below* 🤖

This file implements the same mesh propagation twice: this version constructs and multiplies full `N × N` matrices for every individual component, while `build_unitary_selected_columns_from_components` applies local operations to selected rows. Please retain one shared propagation implementation. Computing the full unitary should be the special case that selects all columns. This would remove substantial duplicate code and avoid the particularly expensive full-matrix multiplication for every phase shifter and beam splitter.

python/mqt/qmap/ph/routing_to_phases.py
```diff
@@ -0,0 +1,171 @@
...
+    Returns:
+        A tuple ``(effective_params, grad_mask, refined_mask)`` where
+        *effective_params* are the physically constrained phase values,
+        *grad_mask* indicates which entries contribute gradients (1.0) or
+        are frozen (0.0), and *refined_mask* is the updated movement mask.
+    """
+    priority_map = {
+        MaskState.MZI: 0,
+        MaskState.BAR: 1,
+        MaskState.CROSS: 2,
```
burgholzer: 🤖 *AI text below* 🤖

This function is called during every optimization iteration, but it traverses the tensor with nested Python loops and repeatedly calls `.item()` on individual entries. Please prepare tensor masks for the fixed routing states outside the optimization loop and apply the bar, cross, and virtual-phase constraints with tensor operations. The current implementation adds Python overhead, and `.item()` can force device synchronization.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
```diff
@@ -0,0 +1,643 @@
...
+            if layer >= start_mzi_layer and top_mode in target_modes and bottom_mode in target_modes:
+                computation_bs_indices.extend((current_bs_idx, current_bs_idx + 1))
+
+            current_bs_idx += 2
+
+    indices_tensor = torch.tensor(computation_bs_indices, dtype=torch.long)
+    return all_bs_values[indices_tensor], indices_tensor
+
+
+@dataclass
```
burgholzer, line 177: 🤖 *AI text below* 🤖

The compiler consumes only `phase_shifter_params` and `best_loss`. The learning-rate history is unused, while the loss and iteration histories appear to be retained only for tests of this private optimizer. Please reduce the result to the values required by the compiler unless these histories are intended to become a supported user-facing API.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
```diff
@@ -0,0 +1,643 @@
...
+    if baseline_outputs is not None:
+        effective_unitary = effective_unitary[baseline_outputs]
+
+    n = effective_unitary.shape[1]
+    overlap = torch.trace(target_unitary.conj().T @ effective_unitary)
+    fidelity = overlap.abs() ** 2 / (n * n)
+    return 1.0 - fidelity
+
+
+def get_computation_zone(
```
burgholzer: 🤖 *AI text below* 🤖

This function has no caller in the package, evaluation, or tests. Please remove it rather than retaining an unused alternative implementation of computation-zone indexing.

python/mqt/qmap/ph/routing.py
```diff
@@ -0,0 +1,337 @@
...
+        A list of length ``chip_dim`` suitable for use as a
+        :class:`perceval.BasicState`.
+    """
+    converted = [0] * chip_dim
+    for port in input_ports:
+        converted[port] = 1
+    return converted
+
+
+def convert_output_ports(output_ports: list[int], chip_dim: int) -> list[int]:
```
burgholzer: 🤖 *AI text below* 🤖

This helper is used only by its own tests. Please remove it. `convert_input_ports` and `get_input_ports_for_computation_zone` also implement the same index-to-binary-vector conversion; if both remain necessary after moving the evaluation, they should share one small helper.

python/mqt/qmap/ph/routing.py
```diff
@@ -0,0 +1,337 @@
...
+        A list of length ``target_dim`` with ``1`` at each active column and
+        ``0`` elsewhere.
+    """
+    result = [0] * target_dim
+    for col in active_columns:
+        result[col] = 1
+    return result
+
+
+def route_to_movement_mask(
```
burgholzer, line 117: 🤖 *AI text below* 🤖

This function independently re-encodes the routing graph topology through separate first-layer, even-layer, and odd-layer transition branches. Please reduce this to one transition rule parameterized by layer parity, or carry the required routing action from graph construction. At present, changes to the mesh topology must be kept consistent in two large implementations.

test/python/ph/test_subcircuit_compilation.py
```diff
@@ -0,0 +1,426 @@
...
+    # variation while still flagging a failure to converge.
+    losses_max: float = field(default=5e-3)
+
+    @property
+    def id(self) -> str:
+        t_tag = "t1.0" if self.t_low is None else f"t{self.t_low:.1f}"
+        return f"chip{self.chip_dim}-target{self.target_dim}-pe{self.phase_error:.3f}-{t_tag}"
+
+
+# fmt: off
```
burgholzer: 🤖 *AI text below* 🤖

This scenario matrix and its calibrated performance bounds are part of the paper evaluation rather than the installed compiler contract. Please move those checks with the evaluation. For the package, retain a small deterministic end-to-end compiler test and focused semantic cases for routing and phase generation. The library tests should permit the internal simplifications requested elsewhere instead of fixing every current helper and representation in place.

burgholzer (changes requested):

burgholzer:
Hey @tobi-forster 👋🏼 

thank you for all the work you have put into this. 
Great to see photonics support in QMAP 😄 

As promised, I started to take a look at this.
However, the PR in its current form is simply too large and mixes too many responsibilities for me to review it properly. It adds close to 6000 lines of code across 30 files and combines the distributed compiler, a Perceval-based evaluation framework, baseline experiments, synthetic hardware generation, data collection, documentation, and a very extensive test suite. A detailed review at this point would spend a lot of time on structure and code that should first be moved, removed, or consolidated. More importantly, it would make it difficult to give the actual compiler algorithm the careful review it deserves.

What I would like to see before I can give this a proper review:
- Keep only the functionality required to compile a subcircuit in the distributed `mqt.qmap.ph` package.
- Move the paper-specific baseline, Perceval simulation, synthetic hardware generation, data collection, notebook, hardware data, and evaluation-specific result types under `eval/ph`.
- Add concise documentation under `eval/ph` that explains the purpose of the evaluation, its dependencies, and how to reproduce it. It's fine if the evaluation code itself does not necessarily meet the same quality standards we impose on the main library.
- Keep the distributed `photonics` extra limited to dependencies needed by the compiler. Evaluation-only dependencies such as pandas and Perceval should not become package dependencies.
- Substantially simplify the compiler implementation. There are unused helpers, a custom shortest-path implementation despite rustworkx already being available, duplicate graph-layer functions, two implementations of the same unitary propagation, and several places that encode the mesh topology independently.
- Refocus the package tests on the public compiler contract and a small number of meaningful semantic cases. Paper performance sweeps and calibrated experimental bounds belong with the evaluation.

I have a hard time prescribing all the necessary steps for the rescoping of this PR in the form of review comments. I used some LLM assistance to identify the main areas of duplication and unnecessary coupling. The attached comments are the result of that.
Take all of them with a grain of salt, but still take them seriously.

I will be happy to review the PR more thoroughly once the PR is refactored accordingly.

eval/ph/data_collection.py
tobi-forster: All evaluation-based code was moved there and a README was added.

pyproject.toml
tobi-forster: Done — mqt.qmap[photonics] now installs only numpy + torch. Pandas (data collection) and Perceval (simulation/evaluation) moved to the non-installed evaluation dependency group under eval/ph, and the associated lint/CI special-cases for the combined set were dropped.

python/mqt/qmap/ph/subcircuit_compilation.py
tobi-forster: Done — both moved to eval/ph/evaluation.py (they're now defined there, alongside evaluate_subcircuit). The installed mqt.qmap.ph API now exposes only CompilationResult (phases + input/output ports to program the hardware); the proposed-vs-baseline comparison structs live entirely in the evaluation code.

python/mqt/qmap/ph/subcircuit_compilation.py
tobi-forster: Done — the whole evaluation path is now evaluate_subcircuit in eval/ph/evaluation.py, together with the baseline optimization, Perceval-based chip construction/simulation, and the comparison metrics. The installed package exposes only compile_subcircuit, so its API and dependency boundary (numpy + torch) no longer pull in Perceval.

python/mqt/qmap/ph/graph.py
tobi-forster: Done — generate_beam_splitter_matrix moved to eval/ph/hardware_model.py. Its module docstring documents the origin and purpose of the hard-coded constants (target_mean = 0.552, target_std = 0.038, target_avg_abs_diff = 0.019 — the fabrication mean/spread and intra-MZI reflectivity difference of the synthetic evaluation chip). The compiler consumes measured reflectivities and never generates them.

python/mqt/qmap/ph/routing.py
tobi-forster: Kept as-is here. The routing search isn't a plain Dijkstra: it's the custom layered shortest-path formulation described in the paper (the mesh topology and per-layer edge semantics are baked into the traversal), so it's part of the photonics-specific routing model rather than a generic graph algorithm we could swap for digraph_dijkstra_shortest_paths. I'd prefer to keep it to better reflect the paper.

python/mqt/qmap/ph/graph.py
tobi-forster: Done — both collapsed into a single get_edge_cost_for_graph_layer in graph.py. It derives the per-layer MZI count from chip-layer parity (mzis_per_even_chip_layer if chip_layer % 2 == 0 else mzis_per_odd_chip_layer) and shares the one validation/offset/fidelity-selection/cost-conversion path.

python/mqt/qmap/ph/graph.py
tobi-forster: Done.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
tobi-forster: Done.

python/mqt/qmap/ph/routing_to_phases.py
tobi-forster: Done.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
tobi-forster: Done — OptimizationResult is reduced to just phase_shifter_params and best_loss. The tests are rewritten accordingly.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
tobi-forster: Done.

python/mqt/qmap/ph/routing.py
tobi-forster: Removed the test-only helper. Also dropped get_input_ports_for_computation_zone. The compiler now works purely on indices, so it's no longer needed. The only remaining conversion, convert_input_ports, is used just for Perceval and now lives in eval/ph/perceval_simulation.py.

python/mqt/qmap/ph/routing.py
tobi-forster: Done — the separate first/even/odd branches are gone, replaced by one parity-based rule (row_start = prev_mode - ((prev_mode - chip_layer % 2) % 2)). The topology is now encoded in one place.

test/python/ph/test_subcircuit_compilation.py
tobi-forster: Done — the scenario matrix and calibrated bounds moved to eval/ph/tests/. The package now has a small deterministic end-to-end test (test/python/ph/test_subcircuit_compilation.py) plus focused routing and phase-generation cases, none of which pin down internal helpers.

noxfile.py
```diff
@@ -83,13 +85,27 @@ def _run_tests(
         *install_args,
         env=env,
     )
+    # === TEMPORARY (macOS-Intel torch wheel gap) - remove once resolved ===
```
burgholzer: 🤖 *AI text below* 🤖

[P2] Now that main no longer supports macOS x86_64, please remove this platform branch and the `platform` import. Pass `optional_dependencies` directly to `uv sync`. Python 3.10 is already outside main's support range, so it needs no compatibility branch here.

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,306 @@
...
+        compilation compute time.
+
+    Note:
+        No hardware simulation is performed, so this step is suitable
+        for chips too large to simulate classically.
+    """
+    if config is None:
+        config = OptimizationConfig()
+
+    chip_dim = len(input_transmissions)
```
burgholzer, line 353: 🤖 *AI text below* 🤖

[P1] Please validate the public compiler inputs before graph construction. The target must be a square, finite complex matrix with a supported even dimension; the output-transmission vector must have exactly `chip_dim` finite values in `[0, 1]`; and an explicit beam-splitter vector must have exactly `chip_dim * (chip_dim - 1)` finite values in `[0, 1]`. The current code silently accepts a short output vector, turns negative transmissions into NaN edge costs, and raises an incidental `IndexError` for a short beam-splitter vector. These cases should raise clear `ValueError`s, with regression tests at this public boundary.

python/mqt/qmap/ph/graph.py
```diff
@@ -0,0 +1,406 @@
...
+                (layers[layer][i], layers[layer + 1][2 * i + 1], cross_costs[i])
+                for i in range(number_nodes_first_layer)
+            ]
+            edges.append(current_edges)
+        elif layer == number_of_layers - 2:
+            edges.append([
+                (layers[layer][i], layers[layer + 1][0], output_transmissions_per_edge[i // 2])
+                for i in range(number_nodes_intermediate_layers)
+            ])
+        elif layer % 2 == 1:
```
burgholzer, line 340: 🤖 *AI text below* 🤖

Ponytail: please collapse the odd/even branches into one pair range and build the bar/cross edges once. In the same construction path, inline the one-use `_edge_cost` wrapper and replace the loop that counts alternating layer widths with a direct sum. These changes remove duplicated topology code without adding another abstraction.

python/mqt/qmap/ph/subcircuit_compilation.py
```diff
@@ -0,0 +1,306 @@
...
+    movement_mask = route_to_movement_mask(best_node_sequence, chip_dim=chip_dim, target_dim=target_dim)
+
+    input_ports, output_ports, active_cols_computation_zone = infer_input_computation_and_output_ports(
+        best_node_sequence, target_dim
+    )
+    _validate_input_ports(input_ports, chip_dim)
+
+    # When photons enter on odd columns (mode 0 is not an active input), apply a swap
+    # permutation to the target so the optimizer sees the correct column ordering.
+    if 0 not in active_cols_computation_zone:
```
burgholzer, line 226: 🤖 *AI text below* 🤖

Ponytail: multiplying by a full permutation matrix is unnecessary here. This operation only swaps each adjacent pair of target columns, so build the column order `[1, 0, 3, 2, ...]` and index `target` directly. That removes the matrix allocation, Python fill loop, and matrix multiplication.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
```diff
@@ -0,0 +1,439 @@
...
+import torch
+
+from .routing_to_phases import get_effective_params_and_mask, reshape_flattened_params_to_grid
+
+logger = logging.getLogger(__name__)
+
+TWO_PI = 2 * torch.pi
+
+
+def build_unitary_from_components(
```
burgholzer: 🤖 *AI text below* 🤖

Ponytail: `build_unitary_from_components` has one caller and always selects all columns, so the caller can invoke the selected builder directly. Please also remove the unused `layer_range` option and its offset loop, and reuse `reshape_flattened_params_to_grid` instead of maintaining the nested `to_grid` copy. These are deletion-only changes within this file.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
```diff
@@ -0,0 +1,439 @@
...
+                phi = ps_grid[num_modes - 1, layer]
+                u = apply_ps_left(u, num_modes - 1, phi)
+
+    return u
+
+
+def fidelity_loss(
+    effective_unitary: torch.Tensor,
+    target_unitary: torch.Tensor,
+    active_cols: list[int] | None = None,
```
burgholzer, line 136: 🤖 *AI text below* 🤖

Ponytail: `active_cols` and `active_cols_target` are dead flexibility. All production callers pass `None` after selecting the target rows or columns earlier. The evaluation-only `baseline` flag also duplicates `output_rows`; the evaluation caller can pass `list(range(target_dim))` instead. Removing these arguments shrinks the optimizer's production contract.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
```diff
@@ -0,0 +1,439 @@
...
+            patience counter.
+
+    Returns:
+        An :class:`OptimizationResult` with the best parameter grid and its loss.
+    """
+    # With a single iteration the loop would evaluate the initial parameters, take one
+    # optimizer step, and terminate before ever evaluating that step's result - leaving
+    # the step wasted and returning the initial parameters. Require at least two
+    # iterations so at least one optimizer step is always evaluated before returning.
+    max_iterations = max(2, max_iterations)
```
burgholzer: 🤖 *AI text below* 🤖

[P2] `max_iterations` is documented as the optimization limit, but clamping it to at least two makes values zero and one run more steps than requested. The loop also evaluates the loss before `optimizer.step()` and returns parameters after the final step without evaluating that state. Please make the iteration count exact, evaluate the returned state consistently, and replace the test that currently pins the clamp with contract-level cases for zero, one, and a small positive limit.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
```diff
@@ -0,0 +1,439 @@
...
+        b = 1j * torch.sqrt(1 - r)
+        row_top = u_local[mode : mode + 1, :]
+        row_bot = u_local[mode + 1 : mode + 2, :]
+        new_top = a * row_top + b * row_bot
+        new_bot = b * row_top + a * row_bot
+        prefix = u_local[:mode, :]
+        suffix = u_local[mode + 2 :, :]
+        return torch.cat((prefix, new_top, new_bot, suffix), dim=0)
+
+    for layer in range(start_layer, end_layer):
```
burgholzer: 🤖 *AI text below* 🤖

[P2] This inner propagation path creates and concatenates full tensors once per component. A 48-mode Clements mesh has roughly 4,500 component operations per loss evaluation, so the default 10,000 optimizer iterations can execute tens of millions of Python-level tensor operations before autograd work. Please batch the independent operations by layer, or provide a representative benchmark that demonstrates the documented mesh sizes are practical. Keep a focused equivalence test for the resulting unitary.

docs/ph_subcircuit_compiler.md
```diff
@@ -0,0 +1,239 @@
...
+physical mode indices — which modes to inject photons into and which modes to
+read out. The router chose these to minimize photon loss.
+
+## The compilation result
+
+{py:class}`~mqt.qmap.ph.subcircuit_compilation.CompilationResult` bundles:
+
+| Field          | Meaning                                                                                               |
+| -------------- | ----------------------------------------------------------------------------------------------------- |
+| `phases`       | `(chip_dim, chip_dim)` tensor of phase-shifter angles to program (rows = modes, columns = MZI layers) |
```
burgholzer: 🤖 *AI text below* 🤖

[P2] This table describes `phases` as a two-dimensional tensor, but the public function and the earlier example return a flat list of length `chip_dim**2`. Please document the actual return shape and keep the terminology consistent across this page.

burgholzer (changes requested):
Hey @tobi-forster 👋🏼 
Thanks for the changes. I had Codex take another spin at this and we are definitely converging towards something mergable!
As pointed out in the message below, I'd leave it up to you whether you want to work through these comments yourself or whether you want me to let get-5.6-sol take a stab add addressing the open comments. 😌 
I am optimistic, that we will get this merged and released before QCE 😌 

🤖 *AI text below* 🤖

Thanks, Tobi. This revision is a clear improvement. Moving the evaluation code to `eval/ph`, narrowing the package boundary, and refocusing the tests addressed most of the original review.

I am keeping the changes-requested review because three earlier points remain open: the custom shortest-path implementation, repeated construction of fixed mask geometry inside the optimizer loop, and duplicated topology rules across the graph, mask, phase-conversion, and unitary paths. I added focused comments for those points.

I also found three approval blockers: the public compiler entry point does not reject malformed dimensions or transmission data, `max_iterations` does not implement its documented contract, and the phase-output documentation disagrees with the API. The propagation loop also needs a practical scaling check before this can support the documented larger meshes.

The separate simplification pass found several deletion-only changes. In particular, main no longer supports x86 macOS, so the Intel-Mac dependency branch and `platform` import can disappear. Main already requires Python 3.11 or newer, and this PR does not need a Python 3.10 compatibility path. The PR description should also stop listing Perceval and pandas as compiler dependencies now that the evaluation code has moved.

All current hosted checks are green. I could not add a trustworthy local compiler run because the selected local environment lacks Torch, so I have not treated that as extra validation.

If it helps, I can have my current Codex session tackle these open points for you. It already has the review context and can prepare a focused patch for you to inspect.

python/mqt/qmap/ph/routing.py
burgholzer: 🤖 *AI text below* 🤖

Thanks for explaining the intent. I rechecked the revised implementation, and the custom traversal does not encode photonic semantics: `_shortest_path` only reads graph layers, edges, and weights. The photonic rules already enter when the graph is constructed.

A native rustworkx Bellman–Ford call therefore preserves those semantics and removes the custom layered relaxation and path reconstruction. I also checked a nontrivial graph against the current implementation and obtained the same route and cost. Dijkstra would require normalizing `-np.log(1.0)` from negative zero, so Bellman–Ford is the smaller direct replacement here. Please replace the custom routine rather than retaining both the graph dependency and a private shortest-path implementation.

python/mqt/qmap/ph/routing_to_phases.py
burgholzer: 🤖 *AI text below* 🤖

The element-wise Python loop is gone, but the fixed work is still rebuilt on every optimizer iteration. `get_effective_transmission_mask_torch` reconstructs mode and layer geometry, partner maps, pair states, and the gradient mask although these values do not change during an optimization run. It also returns an unchanged clone as `refined_mask`, which the callers discard.

Please precompute the static topology once outside the loss loop and keep only the phase-dependent tensor operations in the differentiable path. Remove the unused cloned return value.

python/mqt/qmap/ph/routing.py
burgholzer: 🤖 *AI text below* 🤖

The graph transition rule is more centralized than before, but topology is still encoded independently in `construct_graph`, `get_initial_route_mask`, `routing_to_phase_params`, and the unitary builder. Each path derives odd/even pairing and layer movement separately, so a topology change can still make routing disagree with phase extraction or simulation.

Please make the routed edge or one shared topology representation carry the action needed downstream. At minimum, derive all four paths from one shared pairing rule instead of maintaining parallel parity formulas.

noxfile.py
tobi-forster: Done — removed the Darwin/x86_64 branch and the `platform` import; `optional_dependencies` is passed straight to `uv sync`.

python/mqt/qmap/ph/subcircuit_compilation.py
tobi-forster: Done — `compile_subcircuit` now validates the target (square, even, finite, complex, ≤ chip), both transmission vectors (length `chip_dim`, finite, in [0,1]), and the beam-splitter vector (length `chip_dim*(chip_dim-1)`, finite, in [0,1]), with a regression test per case.

python/mqt/qmap/ph/subcircuit_compilation.py
tobi-forster: Done — replaced the permutation-matrix multiply with direct column indexing (`target[:, [1,0,3,2,…]]`); verified identical for target dims 2/4/6/8.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
tobi-forster: Done — removed the `build_unitary_from_components` wrapper (caller now uses the selected builder with all columns), dropped the unused `layer_range`, and deduped `to_grid` onto `reshape_flattened_params_to_grid`. Verified bit-for-bit across shapes and column selections.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
tobi-forster: Done — removed the floor-of-two clamp so the count is exact, and the loop no longer takes a step whose result isn't evaluated. Replaced the clamp test with contract cases for 0/1/small budgets; the default early-stop path is unchanged.

docs/ph_subcircuit_compiler.md
tobi-forster: Done — the table now documents `phases` as a flat `chip_dim**2` column-major list, matching the API and the example.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
tobi-forster: Not yet done.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
tobi-forster: `active_cols`/`active_cols_target` aren't dead — the compiler passes `active_cols=input_ports, active_cols_target=active_cols_computation_zone` and the eval baseline passes `active_cols=baseline_active_cols`, so they select the sub-block being optimized; left in place. The `baseline` flag is redundant (it only sets `output_rows=list(range(target_dim))`) and can go if you'd like.

python/mqt/qmap/ph/graph.py
tobi-forster: Not yet done.

python/mqt/qmap/ph/routing.py
tobi-forster: Not yet done.

python/mqt/qmap/ph/routing.py
tobi-forster: Done — `get_best_route` now uses `rustworkx.digraph_bellman_ford_shortest_paths`; the custom traversal and path reconstruction are gone. Route and cost are identical to the old implementation on all non-degenerate chips (checked over 600 random chips plus the full eval suite).

python/mqt/qmap/ph/routing_to_phases.py
tobi-forster: Done — split into `precompute_routing_transform` (built once, outside the loop) and `apply_routing_transform` (the only phase-dependent part); removed the unused `refined_mask` return. Optimizer output is unchanged.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
```diff
@@ -0,0 +1,609 @@
...
+        loop_loss = loss.item()
+        losses.append(loop_loss)
+
+        if loop_loss < best_loss - min_improvement:
+            best_loss = loop_loss
+            no_improve_steps = 0
+        else:
+            no_improve_steps += 1
+
+        index % 100 == 0 and verbose
```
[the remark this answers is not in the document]
burgholzer: 🤖 *AI text below* 🤖

Verified in the current code: the optimizer calls `logger.info` every 100 iterations when `verbose=True`, and the docstring describes INFO-level logging. A direct check records iterations 0 and 100 with logging enabled and no calls with it disabled. The discarded Boolean expression is gone.

noxfile.py
burgholzer: 🤖 *AI text below* 🤖

Acknowledged. The current test setup installs the `photonics` extra directly. Paper-reproduction dependencies and tests stay in the separate `evaluation` session. The obsolete macOS x86_64 branch is gone.

python/mqt/qmap/ph/routing.py
burgholzer: 🤖 *AI text below* 🤖

Confirmed Tobi's replacement: routing uses rustworkx Bellman-Ford, with no custom relaxation or predecessor reconstruction. The follow-up also rejects non-finite route costs, so a path containing a zero-transmission edge cannot be returned as a usable route. Fully blocked input and output cases now raise a clear compiler `ValueError`.

python/mqt/qmap/ph/routing_to_phases.py
burgholzer: 🤖 *AI text below* 🤖

Confirmed the precompute/apply split. The follow-up reduces the cached transform to phase sources, constant offsets, and a selector. Each loss evaluation only gathers the source phases and applies the constraints; autograd handles fixed and shared phases without a separate gradient mask. Regression tests cover the routing states and their gradients.

pyproject.toml
burgholzer: 🤖 *AI text below* 🤖

The photonics-specific unresolved-import suppression is now removed. Torch is in the test dependency group, which the default development environment includes, so `ty` checks the compiler and package tests with Torch available. The full lint session passes. The non-installed evaluation remains outside the default type-check scope.

python/mqt/qmap/ph/routing.py
burgholzer: 🤖 *AI text below* 🤖

Implemented in `48680841`: `mzi_top_modes` is now the shared pairing rule for graph construction and costs, movement masks, routing-phase constraints, and unitary propagation. Tests check physical edge costs, reject crosses outside the layer's MZI pairs, and compare propagated unitaries and gradients with an independent component model.

noxfile.py
burgholzer: 🤖 *AI text below* 🤖

Confirmed Tobi's change: the platform branch and import are removed, and `_run_tests` passes `optional_dependencies` directly to `uv sync`. The merge with current `main` retains this setup.

python/mqt/qmap/ph/graph.py
burgholzer: 🤖 *AI text below* 🤖

Implemented in `48680841`: graph construction uses one shared pair range and one bar/cross edge path, and layer widths use a direct sum. The old wrapper is gone. The re-audit also found incorrect MZI offsets and incomplete window costs; each edge now charges every traversed MZI. A focused physical-transition test checks the weights.

python/mqt/qmap/ph/subcircuit_compilation.py
burgholzer: 🤖 *AI text below* 🤖

Confirmed the public-boundary validation and extended its coverage. All three characterization vectors reject non-finite and out-of-range values. The target must be strictly smaller than the chip because this compiler needs routing space. Both `complex64` and `complex128` targets now work; a regression test checks that the returned phases and ports reproduce the reported loss.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
burgholzer: 🤖 *AI text below* 🤖

You are right: my earlier claim about `active_cols` and `active_cols_target` was incorrect. The compiler needs both to map physical input modes to target columns, and the evaluation baseline also selects input columns. Both arguments remain. The redundant `baseline` flag is removed, and the evaluation passes `output_rows` explicitly.

python/mqt/qmap/ph/subcircuit_compilation.py
burgholzer: 🤖 *AI text below* 🤖

Tobi removed the permutation matrix. Tracing the complete flow showed that the remaining pair swap followed by odd-column selection is equivalent to selecting the original target's even columns. The follow-up removes both that swap and the intermediate computation-port list. The compiler now passes `range(0, target_dim, 2)` directly; returned-phase/port tests preserve the physical mapping.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
burgholzer: 🤖 *AI text below* 🤖

Confirmed Tobi's removal of the full-unitary wrapper and `layer_range`. The remaining propagation path now accepts the phase grid directly, so no per-evaluation shape adapter remains. Parameter initialization still uses the existing reshape helper. Selecting all columns and selecting a subset both have unitary and gradient equivalence tests.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
burgholzer: 🤖 *AI text below* 🤖

The final loop evaluates the initial state and each state after a gradient step, then returns the best evaluated phases and matching loss. New tests count actual `Adam.step()` calls for budgets 0, 1, and 3 with early stopping disabled, and independently evaluate the returned state. Negative budgets raise `ValueError`.

docs/ph_subcircuit_compiler.md
burgholzer: 🤖 *AI text below* 🤖

Confirmed: the result table and API docstring describe a flat list of `chip_dim**2` phases in column-major, layer-by-layer order. The example ran successfully. The returned-phase regression reconstructs the grid from that list and checks the reported loss using the returned input and output ports.

python/mqt/qmap/ph/unitary_to_phase_compilation.py
burgholzer: 🤖 *AI text below* 🤖

Implemented layer-batched propagation in `48680841`. Each layer applies its independent beam splitters and phases together, instead of rebuilding a full tensor once per component. An independent component-by-component model checks both unitary values and phase gradients, including selected columns and excluded corner phases.

The local 48-mode, two-column forward-plus-backward benchmark changed from 262.21 ms to 7.36 ms, about 35.6x faster. This measures the propagation kernel, not end-to-end compilation or GPU performance.

burgholzer:
@tobi-forster Alright. As promised, I let Codex run through the remaining open comments here. I resolved all conversations where there was nothing left to be said and left the ones open that had new implementations and may be worth reading from your side. Feel free to resolve them after reading if you are fine with the solution.
I'll monitor CI and see whether any further fixes need to be pushed. 
Otherwise, this should be good to go and we can include it in the QMAP release we are about to release ✨ 🚀 

🤖 *AI text below* 🤖

Pushed the signed commits `48680841` and `580ee7f9`, including the merge of current `main` (`72ee2f2b`).

The main outcome is one shared mesh pairing rule across routing, phase constraints, and propagation, with corrected physical routing costs and much less work in each optimizer iteration.

- Each routing edge now accounts for every traversed MZI. Zero-transmission paths fail with a clear error instead of producing an unusable result.
- Layer-batched propagation preserves unitary values and phase gradients against an independent component model. The local 48-mode, two-column forward/backward benchmark improved from 262.21 ms to 7.36 ms (35.6x); this is a kernel measurement, not an end-to-end speedup.
- The compiler returns the best evaluated, constrained phases and their matching loss. Tests cover both complex dtypes, physical port selection, and exact gradient-step budgets.
- The simplification pass removed redundant gradient masks, compiler wrappers, column permutations, optimizer options, and the NumPy requirement from the compiler extra. The photonics production code has 611 fewer lines than `f2456dea`. The required active-column arguments remain; my earlier recommendation to remove them was incorrect.

Local validation before the push: **1,170 tests passed**, with one unrelated optional-MLIR test skipped; the full `uvx nox -s lint` session, package rebuild, and documentation example passed. One noisy evaluation bound changed from 0.065 to 0.08 because corrected routing selects a different window; an isolated graph-only comparison confirmed the cause.

burgholzer (approved):
Approving this one here now 🚀 
The last step before merging is resolving all comments. I'll leave that to you @tobi-forster, if you feel that all the solutions are fine. Feel free to hit the merge button if everything is green afterwards.

Great to see Photonics in QMAP 🥳

